PR #58, opened 2016-08-26
Add script for making PNGs from CTX and DOS cubes.

### LINKED ISSUE #36: add script: dos+ctx -> png slices
One command in order to plot all the colour wash images.
eog *.jpg can then be used to flip through the slices.

nbassler:
Resolves #36 

This can be extended and improved indefinitly, for now I just wanted something which works.

### DIFF
--- a/pytrip/utils/slicedos.py
+++ b/pytrip/utils/slicedos.py
@@ -2,16 +2,18 @@ from pytrip import dos
 from pytrip import ctx
 import argparse
 import logging
-from numpy import arange,dtype,meshgrid
+from numpy import arange, dtype, meshgrid, ma
 import matplotlib.pyplot as plt
+from matplotlib import colors
+import os
 
 
 def check_compatible(a, b):
-    """ Simple comparison of cubes. if X,Y,Z dims are the same, and 
+    """ Simple comparison of cubes. if X,Y,Z dims are the same, and
     pixel sizes as well, then they are compatible. (Duck typed)
     """
     eps = 1e-5
-    
+
     if a.dimx != b.dimx:
         logging.error("DIMX does not match: "+str(a.dimx)+" "+str(b.dimx))
         raise Exception("Cubes don't match, check dimx in header.")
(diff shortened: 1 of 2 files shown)


grzanka:
I recommend to put contents of the script into main function, like here:
https://github.com/pytrip/pytrip/blob/master/pytrip/utils/trip2dicom.py

This allows to make tests, like these:
https://github.com/pytrip/pytrip/blob/master/tests/test_converters.py

Moreover setup.py should also be extended:
https://github.com/pytrip/pytrip/blob/master/setup.py#L167
If the script is gonna be called by user it should have a name.

When users installs pytrip via pip, then all entry points will be added to $PATH.

nbassler:
done.

pytrip/utils/slicedos.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def check_compatible(a, b):
+    """ Simple comparison of cubes. if X,Y,Z dims are the same, and
+    pixel sizes as well, then they are compatible. (Duck typed)
+    """
+    eps = 1e-5
+
+    if a.dimx != b.dimx:
+        logging.error("DIMX does not match: "+str(a.dimx)+" "+str(b.dimx))
```
grzanka: Better to create logger instance (`logging.getLogger`) instead of using module methods, see: http://victorlin.me/posts/2012/08/26/good-logging-practice-in-python

pytrip/utils/slicedos.py
```diff
@@ -0,0 +1,211 @@
...
+        logging.error("Pixel size does not match: "+str(a.pixel_size)+" "+str(b.pixel_size))
+        raise Exception("Cubes don't match, check pixel_size in header.")
+
+    if a.slice_dimension != b.slice_dimension:
+        logging.error("Slice dimension does not match: "+str(a.slice_dimension)+" "+str(b.slice_dimension))
+        raise Exception("Cubes don't match, check slice_dimension in header.")
+    return True
+
+
+def main():
```
grzanka: To have working tests you need to have `main` with arguments, see:
https://github.com/pytrip/pytrip/blob/master/pytrip/utils/trip2dicom.py#L9
`def main(args=sys.argv[1:]):`
and pass these aguments to `argparse.ArgumentParser`

https://github.com/DataMedSci/pymchelper/blob/master/pymchelper/run.py#L52
` parsed_args = parser.parse_args(args)`

Having args enable you to use them in tests, see:
https://github.com/pytrip/pytrip/blob/master/tests/test_converters.py#L21
` pytrip.utils.trip2dicom.main([self.cube000, tmpdir])`

pytrip/utils/slicedos.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # majorLocator = plt.MultipleLocator(1)
+        # majorFormatter = plt.FormatStrFormatter('%d')
+        minorLocator = plt.MultipleLocator(1.5)
+
+        ax.xaxis.set_minor_locator(minorLocator)
+        plt.savefig("foo"+str(ids)+".png")
+
+if __name__ == '__main__':
+    sys.exit(main())
```
grzanka: Args should also go here, like `    sys.exit(main(sys.argv[1:]))`

pytrip/utils/slicedos.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # extent tells what the x and y coordinates are,
+        # so matplotlib can make propper assignment.
+        ctx_im = ax.imshow(ctx_slice,
+                           cmap=plt.cm.gray,
+                           interpolation='bilinear',
+                           origin="lower",
+                           extent=[Xmin, Xmax,
+                                   Ymin, Ymax])
+        if args.HUbar is True:
```
grzanka: ` if args.HUbar is True:` can be replaced with  `if args.HUbar:`

nbassler:
should also add the GPL banner

grzanka:
I'd recommend to do only small fixes here and merge this PR quickly.
Then a new issue should be created to add unit tests which will ensure that dosslicer works correctly.
These test can use data sample provided in PR #59

nbassler:
@grzanka What do you think about the last commit, from googling I have an impression this is more standard way to do it.

grzanka:
Last commit looks OK. I agree - it makes more sense to do it like this, insted of putting parsing inside main function.

pytrip/utils/slicedos.py
```diff
@@ -0,0 +1,213 @@
...
+    if args.verbosity > 1:
+        logger.basicConfig(level=logging.DEBUG)
+
+    logger.info("Dos file: " + args.dos)
+
+    dosbasename = args.dos.split(".")[-2]
+    ctxbasename = args.ctx.split(".")[-2]
+
+    d = dos.DosCube()
+    fndos = dosbasename + ".dos"
```
grzanka: For a future reference: ".dos" can be obtained from a `data_file_extension` field in `DosCube` class.

Issue #40, opened 2016-08-23
One of examples at https://neptun.phys.au.dk/~bassler/TRiP is empty


grzanka:
It looks that one of examples mentioned here https://neptun.phys.au.dk/~bassler/TRiP/ is empty.
Its description says 

> - TST003 : Water cube with box target

Inspected with hex editor (no duplicates are printed):
```
→ od -t x1 -An tests/TST003/tst003000.ctx
 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
*
```

Read with pytrip:
```
# grzanka at gajewski-dell in ~/PycharmProjects/pytrip on git:master ✖︎ [17:57:03]
→ cat tests/bug.py                       
from pytrip import CtxCube

c = CtxCube()
c.read("tests/TST003/tst003000.ctx")

print("shape", c.cube.shape)
print("element [0][0][0]", c.cube[0][0][0])
print("min value", c.cube.min())
print("max value", c.cube.max())

# grzanka at gajewski-dell in ~/PycharmProjects/pytrip on git:master ✖︎ [17:57:06]
→ PYTHONPATH=. python3 tests/bug.py
shape (300, 512, 512)
element [0][0][0] 0
min value 0
max value 0
```

grzanka:
@nbassler can you take a look ?

grzanka:
TST0002 (referred as `TST002 : Water cube with spherical target`) seems also to have empty CT cube:

```
# grzanka at gajewski-dell in ~/PycharmProjects/pytrip on git:master ✖︎ [18:21:50]
→ cat tests/bug.py                                            
import sys
from pytrip import CtxCube

c = CtxCube()
c.read(sys.argv[1])

print("shape", c.cube.shape)
print("element [0][0][0]", c.cube[0][0][0])
print("min value", c.cube.min())
print("max value", c.cube.max())

# grzanka at gajewski-dell in ~/PycharmProjects/pytrip on git:master ✖︎ [18:21:55]
→ PYTHONPATH=. python3 tests/bug.py tests/TST002/tst002000.ctx
shape (300, 512, 512)
element [0][0][0] 0
min value 0
max value 0

```

nbassler:
Water has HU 0?

grzanka:
Me stupid, you are right.

But on the other hand - does it make sense to have different .ctx and .hed files in examples TST001-TST003 ?

nbassler:
[...]

PR #59, opened 2016-08-26
gzip support

### LINKED ISSUE #52: Add reading gzipped .ctx files
This will end discussion in #40

nbassler:
Should close #52 and #40 

Also removed dependency on byteswap() function when reading cubes.


nbassler:
Btw, .hed.gz are also allowed.

pytrip/cube.py
```diff
@@ -402,29 +405,106 @@ def read(self, path):
...
+        if f_split[1] == ".gz":
+            is_zipped = True
+            f_split = os.path.splitext(f_split[0])  # get rid of .gz suffix
+        # now proper header file basename is set, but without .gz suffix.
         header_file = f_split[0] + ".hed"
+
+        # Does the .hed file exist? If not retry with .hed.gz suffix
+        # PyTRiP will also attempt to load .hed.gz files if user did
+        # not specify the .gz suffix explicitly
+        if os.path.isfile(header_file) is False:
```
grzanka: ` if os.path.isfile(header_file) is False:` can be replaced with  `if not os.path.isfile(header_file)`:

grzanka:
I am trying to improve the code to handle situation when we have to read a couple: tst003000.ctx.gz + tst003000.hed. Right now trip2dicom has problem with it.

PR #79, opened 2016-09-09
Dosslice refactoring

### LINKED ISSUE #64: Refactor dosslicer script
- [ ] avoid hardcoding data file extensions
- [ ] add default value for `sstop`
- [ ] ctx cube should be optional 
- [ ] move `check_compatible` method to Cube class
- [x] think if foo is the best name for output file

grzanka:
Fixes #64


grzanka:
@nbassler ready to be merged.

PR #127, opened 2016-10-12
TripExec example code

grzanka:


examples/trip98.py
```diff
@@ -0,0 +1,72 @@
...
+my_ctx.read(ctx_path)
+
+# load VOIs
+my_vdx = pt.VdxCube("", my_ctx)  # my_vdx is the object which will hold all volumes of interest and the meta information
+my_vdx.read(vdx_path)  # load the .vdx file
+print(my_vdx.get_voi_names())  # show us all VOIs found in the .vdx file
+
+# next pick a the proper VOI which we want to plan on
+target_voi_temp = my_vdx.get_voi_by_name(my_target_voi)  # Select the requested VOI from the VdxCube object
+target_voi = pte.Voi("GTV_VOI", target_voi_temp)  # for technical reasons the voi must be cast into a new Voi object
```
nbassler: Think we should add some comment that we know this is not good, and we hope to have some resolution for this in a future release.

nbassler:
Can you also add the vdx example you did?

nbassler:
```
This example demonstrates how to load a CT cube in Voxelplan format, and the associated contours. Then a plan is prepared and optimized using TRiP98. The resulting dose plan is then stored in <some_dir>.
```

PR #134, opened 2016-10-17
vdx example code

### LINKED ISSUE #132: include vdx example
simple example which:
- loads a CTX cube
- loads a VDX file
- print a list of all available contours
- selects by mask one contour
- sets all HU units to zero
- saves new CTX cube to file

grzanka:
Fixes #132

### DIFF
--- a/examples/example01_vdx.py
+++ b/examples/example01_vdx.py
@@ -3,8 +3,8 @@ from pytrip.utils import cubeslice
 
 # first define some paths and other important parameters
 patient_name = "TST000000"
-ctx_path = "absolute_path_to_some_directory/TST000000.ctx"
-vdx_path = "absolute_path_to_some_directory/TST000000.vdx"
+ctx_path = "/home/bassler/Projects/CTdata/TST000/TST000000.ctx"
+vdx_path = "/home/bassler/Projects/CTdata/TST000/TST000000.vdx"
 my_target_voi = "GTV"
 
 # load CT cube
--- a/examples/example01_vdx.py
+++ b/examples/example01_vdx.py
@@ -1,0 +1,38 @@
+import pytrip as pt
+from pytrip.utils import cubeslice
+
+# first define some paths and other important parameters
+patient_name = "TST000000"
+ctx_path = "absolute_path_to_some_directory/TST000000.ctx"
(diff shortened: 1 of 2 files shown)


examples/example01_vdx.py
```diff
@@ -0,0 +1,38 @@
...
+my_vdx.read(vdx_path)  # load the .vdx file
+print(my_vdx.get_voi_names())  # show us all VOIs found in the .vdx file
+
+# next pick a the proper VOI which we want to plan on
+target_voi = my_vdx.get_voi_by_name(my_target_voi)  # Select the requested VOI from the VdxCube object
+
+# dos cube representing VOI, it is filled with zeros outside the volume and with 1000 inside
+voi_cube = target_voi.get_voi_cube()
+
+# mask object, same shape as dos and voi cube, true inside volume, false outside
```
nbassler: "mask object" could also be understood as "to mask the object" as "mask" can both be a verb or a noun. Propose: 

"We transform the retrieved DosCube() to a three dimensional mask object, which assigns `True` to all voxels inside the Voi, and `False` elsewhere."

examples/example01_vdx.py
```diff
@@ -0,0 +1,38 @@
...
+mask = (voi_cube.cube == 1000)
+
+# make a masked CT - set Hounsfield units to zeros inside the volume
+my_ctx.cube[mask] = 0
```
nbassler: "The mask object and the CTX cube have similar dimensions (they are infact inherited from the same top level class). Therefore we can apply the mask cube to the ctx cube and work with the values. For instance we can set all HUs to zero within the Voi:"
# my_ctx.cube[mask] = 0
# or set it to some special based on an algorithm
# my_ctx_cube[mask] += 10  # increase HUs by 10
my_ctx.cube[mask] = <we need some index dependent operation here> # add random noise to the Voi

examples/example01_vdx.py
```diff
@@ -0,0 +1,38 @@
...
+
+# save masked CT to the file in current directory
+masked_ctx = "masked.ctx"
+my_ctx.write(masked_ctx)
+
+# optionally save one slice from original and masked CTs to the png images in current directory
```
nbassler: But is this needed? We have nice pytripgui for viewing the output :grin:

Issue #66, opened 2016-08-28
Investigate header file extensions


grzanka:
See discussion here: https://github.com/pytrip/pytrip/pull/59#commitcomment-18805217

By looking at https://github.com/pytrip/pytrip/tree/master/tests/res/TST003 it is hard to say what is the core name and extensions for LetCube:

Which one is correct ?
- basename `tst003001.dosemlet`, header extension `.hed`, data file extension `.dos`
- basename `tst003001`, header extension `.dosemlet.hed`, data file extension `.dosemlet.dos`

@nbassler do you have an opinion ?

nbassler:
From TRiP point of view, it should be basename `tst003001`, and the file extension is `.dosemlet.hed`. and `.dosemlet.dos`. LET cubes should also be handled differently than dos cubes when plotting, and the only clear evidence that this is a LET cube is the fileextension (and guessing by looking at the data). AFAIK there is no hint in the .hed file whether this is a dose or LET cube.

nbassler:
@grzanka this one may be closed?

grzanka:
@nbassler not yet

nbassler:
@grzanka I am not sure why this is still open?

grzanka:
It is open because current implementation (v1.2.2) has extension `.dos` for LET data cube file. See:
https://github.com/pytrip/pytrip/blob/v1.2.2/pytrip/let.py#L38

```
class LETCube(Cube):
(...)
    data_file_extension = "dos"
```

This also means that currently `parse_path` method of Cube class: https://github.com/pytrip/pytrip/blob/master/pytrip/cube.py#L499
thinks that file with `.dos` extension can be either dose cube or let cube.

Also right now all header files have univeral extension `.hed`

We also assume currently that `filename = core_name + extension` and when reading a cube we need two files:
 - `corename +'.hed'`
 - `corename + data_file_extension`

This means we can read for example pair of file `1.hed` + `1.ctx`, but we do not support reading `2.hed` + `1.ctx`. Both files currently needs to have same corename.

To fix this issue following things has to implemented:
 - change data_file_extension for LETCube to `dosemlet.dos`
 - introduce header extension specific to a data class: `dosemlet.hed` for LETCube and `.hed` for all other classes

nbassler:
this sounds correct to me.

nbassler:
- [ ] reminder to also fix cubeslice TODO

PR #203, opened 2016-12-19
Fix #199, dosecubes not being recognized.

### LINKED ISSUE #199: cubeslice thinks dose cubes are LET cubes
and plots a LET bar, even if the data are dose.

nbassler:
Tested with TST001 from neptun.phys.au.dk/TRiP

Closes #199

### DIFF
--- a/pytrip/utils/cubeslice.py
+++ b/pytrip/utils/cubeslice.py
@@ -47,8 +47,9 @@ def load_data_cube(filename):
     d = None
     basename_cube = None
 
+    # TODO to be improved when #66 is fixed"
     if "dosemlet" in filename.lower():
-        # try to load LET cube    
+        # try to load LET cube
         data_header, _ = pt.LETCube.parse_path(filename)
         if data_header is not None:
             basename_cube = os.path.splitext(data_header)[0]
--- a/pytrip/utils/cubeslice.py
+++ b/pytrip/utils/cubeslice.py
@@ -47,17 +47,18 @@ def load_data_cube(filename):
     d = None
     basename_cube = None
 
-    # try to load LET cube
-    data_header, _ = pt.LETCube.parse_path(filename)
-    if data_header is not None:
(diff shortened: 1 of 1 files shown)


pytrip/utils/cubeslice.py
```diff
@@ -47,17 +47,18 @@ def load_data_cube(filename):
...
-    if data_header is not None:
-        basename_cube = os.path.splitext(data_header)[0]
-        d = pt.LETCube()
-
-    # try to load DOS cube
-    data_header, _ = pt.DosCube.parse_path(filename)
-    if d is None and data_header is not None:
-        basename_cube = os.path.splitext(data_header)[0]
-        d = pt.DosCube()
+    if "dosemlet" in filename.lower():
```
grzanka: What about fixing also (first?) https://github.com/pytrip/pytrip/issues/66 ?

extension is stored in pytrip, see https://github.com/pytrip/pytrip/blob/master/pytrip/let.py#L38

pytrip/utils/cubeslice.py
```diff
@@ -47,17 +47,18 @@ def load_data_cube(filename):
...
+        # try to load LET cube    
+        data_header, _ = pt.LETCube.parse_path(filename)
```
grzanka: parse_path takes care about extension, see docstring here:
https://github.com/pytrip/pytrip/blob/master/pytrip/cube.py#L499

but the extension has to be set correctly.

pytrip/utils/cubeslice.py
nbassler: I have no solution to #66 other that what I just did here.

pytrip/utils/cubeslice.py
nbassler: hey leszek, I dont think we are time effective now, if you can you may directly edit the branch as you please.

nbassler: done

PR #204, opened 2016-12-19
Fix bad slice numbering

### LINKED ISSUE #200: cubeslice from-to behaviour is not as expected
- would expect [FROM:TO] but if FROM==TO then no output is made. if they differ then [FROM:TO[
- also counting the slices outputted starts at zero (0/102) and terminates one index earlier than expected (101/102)

nbassler:
Closes #200

### DIFF
--- a/pytrip/utils/cubeslice.py
+++ b/pytrip/utils/cubeslice.py
@@ -123,7 +123,7 @@ def main(args=sys.argv[1:]):
     parser.add_argument("--data", help="data cube(dos, let etc)", type=str, nargs='?')
     parser.add_argument("--ct", help="CT cube", type=str, nargs='?')
     parser.add_argument(
-        "-f", "--from", type=int, dest='sstart', metavar='N', help="Output from slice number N", default=0)
+        "-f", "--from", type=int, dest='sstart', metavar='N', help="Output from slice number N", default=1)
     parser.add_argument("-t", "--to", type=int, dest='sstop', metavar='M', help="Output up to slice number M")
     parser.add_argument("-H", "--HUbar", dest='HUbar', default=False, action='store_true', help="Add HU colour bar")
     parser.add_argument("-o", "--outputdir", dest='outputdir',
@@ -200,6 +200,10 @@ def main(args=sys.argv[1:]):
     ct_im = None
     data_im = None
 
+    if args.sstart == 0:
+        logger.error("First slice must be 1 or greater.")
+        exit(1)
+        
     # if user hasn't provided number of first slice, then argument parsing method will set it to zero
     slice_start = args.sstart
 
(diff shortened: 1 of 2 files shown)


pytrip/utils/cubeslice.py
```diff
@@ -200,6 +200,10 @@ def main(args=sys.argv[1:]):
     ct_im = None
     data_im = None
 
+    if args.sstart == 0:
```
grzanka: Isn't it better to check if `args.sstart < 1` ?

pytrip/utils/cubeslice.py
```diff
@@ -200,6 +200,10 @@ def main(args=sys.argv[1:]):
...
+        logger.error("First slice must be 1 or greater.")
```
grzanka: I'd recommend also to provide an information about value provided by user, i.e.:
"you provided XXXX, but first slice must be 1 or greater"

pytrip/utils/cubeslice.py
```diff
@@ -227,20 +231,31 @@ def main(args=sys.argv[1:]):
...
             # use os.path.join() in order to add slash if it is missing.
             output_filename = os.path.join(args.outputdir, os.path.basename(output_filename))
         if args.verbosity == 0:
-            print("Write slice number: " + str(slice_id) + "/" + str(cube.dimz))
+            print("Write slice number: " + slice_str)
         if args.verbosity > 0:
-            logger.info("Write slice number: " + str(slice_id) + "/" + str(cube.dimz) + " to " + output_filename)
+            logger.info("Write slice number: " + slice_str + " to " + output_filename)
+
+        if ct_cube is not None:
```
grzanka: You can reduce this code to:
```slice_str += " zpos = {:.2f} mm".format(cube.slice_to_z(slice_id - 1))```
We have `cube` variable which points to `data_cube` or, if its missing to `ct_cube`.
See line 168

File: pytrip/utils/cubeslice.py [hunk 4]
```diff
@@ -227,20 +231,31 @@ def main(args=sys.argv[1:]):
     ax.xaxis.set_minor_locator(minorLocator)
     ax.yaxis.set_minor_locator(minorLocator)
 
+    text_slice = ax.text(0.0, 1.05, "", size=8, transform=ax.transAxes)
```
grzanka: Why not to use plot title to hold such information ?

nbassler: I might want to move it later if more info is added. In fact id like smth like the pytripgui rendering. But this really not important right now..

grzanka:
Here https://github.com/pytrip/pytrip/blob/master/tests/test_converters.py#L269 test expects 1 file, but after your change two are generate. Fix test please

Issue #198, opened 2016-12-18
cubeslice broken


nbassler:
```
bassler@grgalt:~/Projects/shieldhit/test/26_voxel$ cubeslice --data myvox02.dos --ct ../../res/TST000/TST000000.ctx 
Traceback (most recent call last):
  File "/home/bassler/.local/bin/cubeslice", line 11, in <module>
    sys.exit(main())
  File "/home/bassler/.local/lib/python2.7/site-packages/pytrip/utils/cubeslice.py", line 158, in main
    logger.info("Cube 1 " + args.dos + " shape {:d} x {:d} x {:d}".format(data_cube.dimx, data_cube.dimy,
AttributeError: 'Namespace' object has no attribute 'dos'
bassler@grgalt:~/Projects/shieldhit/test/26_voxel$ 
```

nbassler:
but it works if I leave the --ct argument away.

grzanka:
which version of pytrip ? Can you type `cubeslice --version` ?

nbassler:
1.2.2

Not sure what is going on, it seems to work if the filenames are identical or so, maybe related to unsyncronized headers or so. Anyway, a clearer error message would be helpful.

grzanka:
Can you put here header of myvox02.dos ? Is it present in same directory as myvox02.dos ?

nbassler:
working header: TST000000.hed (for .dos generated with TRiP)
```
version 1.2
modality CT
primary_view transversal
data_type integer
num_bytes 2
byte_order vms
patient_name TST000
slice_dimension 256
pixel_size 1.210938
slice_distance 3
slice_number 102
xoffset 0
dimx 256
yoffset 0
dimy 256
zoffset 0
dimz 102
```

Broken header TST000001.hed (and .dos generated from pymchelper ``` convertmc tripcube```
```
version 2.0
modality CT
created_by pytrip
creation_info create by pytrip
primary_view transversal
data_type integer
num_bytes 2
byte_order vms
patient_name Anonymeous
slice_dimension 102
pixel_size 0.121093802154
slice_distance -0.303921699524
slice_number 102
xoffset 0
dimx 256
yoffset 0
dimy 256
zoffset 0
dimz 102
z_table no
```

maybe mostly a pymchelper issue. Cubeslice could throw an error if something bogus is found.

grzanka:
I agree that the error handling should be improved.  Cubeslice expects file `myvox02.hed` - is it present ?

[...]

PR #213, opened 2016-12-28
Improved reading of LET and DOS cubes in cubeslicer

### LINKED ISSUE #212: Improve handling of LET cubes in cubeslicer
Following https://github.com/pytrip/pytrip/issues/66#issuecomment-268022764

grzanka:
now properly using filename parsing
improve debugging in #198
fixes #212

### DIFF
--- a/pytrip/utils/cubeslice.py
+++ b/pytrip/utils/cubeslice.py
@@ -47,21 +47,27 @@ def load_data_cube(filename):
     d = None
     basename_cube = None
 
-    # TODO to be improved when #66 is fixed"
-    if "dosemlet" in filename.lower():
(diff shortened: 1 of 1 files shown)


grzanka:
Until we don't get rid of them - yes. I can think of something smarter which would allow slicing of:
 - only dose
 - only let
 - only CT
 - CT + LET
 - CT + dose

Handling these combinations with `--` options is easy. 
I do not see an easy option to do the same with regular positional arguments.

nbassler:
ok, that is a new issue

grzanka:
After merging this PR, can you check if #198 is still valid ?

PR #231, opened 2017-02-03
Add support for ".IMA" dicom file extension

### LINKED ISSUE #30: DICOM: extend list of possible file extensions
.dcm and .ima are known, plus their upper/lower cases. 
Possibly better to leave extensions out all together, and let the parser try to figure out if the file is valid.

nbassler:
closes #30

### DIFF
--- a/pytrip/dicomhelper.py
+++ b/pytrip/dicomhelper.py
@@ -33,7 +33,7 @@ def read_dicom_folder(path):
 
     # list of allowed dicom file extentions names
     # all in lower case
-    dicom_suffix = ['.dcm','.ima']
+    dicom_suffix = ['.dcm', '.ima']
 
     data = {}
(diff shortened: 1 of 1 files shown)


pytrip/dicomhelper.py
```diff
@@ -30,10 +30,15 @@ def compare_dicom_key(dcm):
 def read_dicom_folder(path):
     if not os.path.isdir(path):
         raise IOError("Folder does not exist")
+
+    # list of allowed dicom file extentions names
+    # all in lower case
+    dicom_suffix = ['.dcm', '.ima', '.v2']
```
grzanka: tuple is better than list if you want to store array of read-only data

grzanka: anyway, its not critical

Issue #196, opened 2016-12-13
.rst class could go into pytrip core


nbassler:
currently it is totally independent script. 

User may want to work with RST files directly, so it should be able to import the classes related to this.

Issue #97, opened 2016-09-30
api change in vdx.py


nbassler:
These issues should be fixed, but will break compability to pytripgui.
- [x] VdxCube.**init** takes "content" as argument, but it is not used, and also not logical. Should be removed.
- [ ] VdxCube class does not make sense, it is no cube. It should just be called Vdx
- [ ] VoiCube class is the same issue. It should be called Voi instead

nbassler:
Recommend to rewrite vdx.py. We should define the API and the structure first, then code it again.

nbassler:
Renaming VdxCube to Vdx will be done for new Vdx reader, see #197 

Closing this one, as the first part was fixed in #234

Issue #96, opened 2016-09-30
VdxCube() seems to be broken


nbassler:
Python3:
```
>>> v.read("TST000000.vdx")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/pytrip/vdx.py", line 95, in read
    self.read_vdx(path)
  File "/usr/local/lib/python3.5/dist-packages/pytrip/vdx.py", line 121, in read_vdx
    i = v.read_vdx(content, i)
  File "/usr/local/lib/python3.5/dist-packages/pytrip/vdx.py", line 498, in read_vdx
    self.name = string.join(line.split()[1:], ' ')
AttributeError: module 'string' has no attribute 'join'
```

Python2:
```
>>> v.read("TST000000.vdx")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/pytrip/vdx.py", line 95, in read
    self.read_vdx(path)
  File "/usr/local/lib/python2.7/dist-packages/pytrip/vdx.py", line 121, in read_vdx
    i = v.read_vdx(content, i)
  File "/usr/local/lib/python2.7/dist-packages/pytrip/vdx.py", line 512, in read_vdx
    key = int((float(s.get_position()) - min(self.cube.slice_pos)) * 100)
AttributeError: 'NoneType' object has no attribute 'slice_pos'
```

nbassler:
These issues can probably be fixed with a dirty hack, however on the long run I recommend to rewrite this part to something more clean.  See #97

grzanka:
I've tested with this file: https://github.com/pytrip/pytrip/blob/master/tests/res/TST003/tst003000.vdx
and it seems to work:

```
→ python3.4
Python 3.4.3 (default, Sep 14 2016, 12:36:27) 
[GCC 4.8.4] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pytrip
>>> from pytrip.vdx import VdxCube
>>> v = VdxCube("")
>>> v.read("tests/res/TST003/tst003000.vdx")
['voi', 'target', 'type', '1', '#subvoi', '1']
>>> 
```

@nbassler  can you reproduce the same problem with tests/res/TST003/tst003000.vdx file ?
From your message its not clear what is v object. Can you copy-paste full session, from starting python interpreter ?

grzanka:
Are there two different formats of VDX files ?

[...]

PR #234, opened 2017-02-04
Unused content parameter removed

### LINKED ISSUE #190: VdxCube constructor gets a parameter which is not used
Vdx constructor gets parameter called `content` which is not used, see https://github.com/pytrip/pytrip/blob/v1.2.1/pytrip/vdx.py#L77

```    def __init__(self, content, cube=None):
        self.vois = []
        self.cube = cube
        self.version = "1.2"
```

This leads to buggy usages, like this one: https://github.com/pytrip/pytrip/blob/v1.2.1/pytrip/utils/trip2dicom.py#L62

```
        v = pt.VdxCube(c)
```

Here `c` is assigned to `content` and not to `cube` as intended

grzanka:
Fixes #190

### DIFF
--- a/examples/example01_vdx.py
+++ b/examples/example01_vdx.py
@@ -31,7 +31,7 @@ my_ctx = pt.CtxCube()
 my_ctx.read(ctx_path)
 
 # load VOIs
-my_vdx = pt.VdxCube("", my_ctx)  # my_vdx is the object which will hold all volumes of interest and the meta information
+my_vdx = pt.VdxCube(my_ctx)  # my_vdx is the object which will hold all volumes of interest and the meta information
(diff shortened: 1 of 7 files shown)


nbassler:
I hope it wont break pytripgui.

nbassler:
Will also fix #97

grzanka:
I will check pytripgui before making new release

grzanka:
pytripgui also needs to be fixed, see https://github.com/pytrip/pytripgui/blob/83a1081b3b13bc4d9386d4cbc50ccd7dbc28cded/pytripgui/data.py#L167

nbassler:
Closes #196 
Fixes one of the three items in #96

PR #235, opened 2017-02-04
Attempt to increase code coverage

nbassler:


tests/test_rst.py
```diff
@@ -17,86 +17,66 @@
...
+        self.rst000 = os.path.join(testdir, "tst003001.rst")
+        logger.info("Testing .rst from path " + self.rst000)
+
+    def test_read(self):
+        """ Check if we are able to read a simple .rst file
+        """
+        r = Rst()
+        r.read(self.rst000)
+        self.assertEqual(r.submachines, '17')
+        self.assertEqual(r.machines[0].points, 323)
```
grzanka: Looks good, but before checking what is under `r.machines[0].points` it would be good to check if length of `r.machines` list is > 0.

nbassler: Why? If it is =0 it will fail... :p

grzanka: Yes, it will fail, but the reason won't be clear.

nbassler: better would be to compare int(r.submachines) to len(r.machines), they should be equal. But there are 100x more of these comparisons which can be included. 
Feel free to do it, this is good enough to me for now.

grzanka:
I will merge as it is in order to get today release 1.3.0

nbassler:
But codecov said it did not change anything? Wtf?

grzanka:
I've just realised that there was already test for RST, in `test_converters.py` file. This is the reason why adding your test suite haven't changed anything

nbassler:
The converters used the old script, but probably it changed when I updated rst2sobp to the new class. Meh.

PR #224, opened 2017-02-02
Support handling bogus pinnacle dicom tags.

nbassler:
Closes #223

PR #236, opened 2017-02-07
resolve UID issues

### LINKED ISSUE #191: DICOM export unique UID
Eclipse (and probably other TPSes) assumes that every new CT scan and every new VOI DICOM has new UID. After opening any DICOM in Eclipse it loads it and keeps it in its database for future reference.

Right now pyTRiP98 saves fixed UIDs to exported DICOMs. It means if you load in Eclipse DICOM file exported from pyTRiP then first file will load correctly. But when you take i.e. different CT cube, export it to DICOM then Eclipse import will fail - Eclipse will complain that DICOM with same UID already exists.

The way to solve that problem would be to generate unique UIDs, based for example on hashes from CTX and HED data. I think we should still assume that two identical CTX+HED sets leads to the same UIDs - generating UIDs based on timestamps doesn't make sense.

### LINKED ISSUE #193: DICOM: long list of warnings when Eclipse is loading PyTRiP generated data
(scroll to the right to see warnings)

missing attributes:

- [x] 'StudyID' (0020,0010)
- [x] 'ReferringPhysiciansName' (0008,0090)
- [x] 'PositionReferenceIndicator' (0020,1040)
- [x] 'SeriesNumber' (0020,0011)
- [x] 'KVP' (0018,0060)
- [x] 'AcquisitionNumber' (0020,0012)

```
Date                 Object                                                                                     Level      Category        Log Message
===================================================================================================================================================================================
11/27/2016 22:00:37  General                                                                                    Info       GENERAL         ReadFile                 : Reading DICOM Media File <C:\Users\IMPS\Documents\example_data\out\ct.0.dcm>
11/27/2016 22:00:37  2.16.1.113662.2.12.0.3057.1241703565.10                                                    Warning    Reading         |W| ImportId_CRmCTmRDmRImRSmREGmDREGmTPmTWmMRmPTmSCmRPmRNmRTmRXmUSm: Type 2 Attribute 'StudyID' (0020,0010) is missing in DICOM stream
11/27/2016 22:00:37  2.16.1.113662.2.12.0.3057.1241703565.10                                                    Warning    Reading         |W| ImportReferringPhysiciansName_CRmCTmRDmRImRSmREGmDREGmTPmTWmMRmPTmSCmRPmRNmRTmRXmUSm: Type 2 Attribute 'ReferringPhysiciansName' (0008,0090) is missing in DICOM stream
11/27/2016 22:00:37  2.16.1.113662.2.12.0.3057.1241703565.10                                                    Warning    Reading         |W| ImportPositionReferenceIndicator_CTmRDmRIuREGmDREGmTPuTWuMRmPTmRPuRNmUSm: Type 2 Attribute 'PositionReferenceIndicator' (0020,1040) is missing in DICOM stream
11/27/2016 22:00:37  2.16.1.113662.2.12.0.3057.1241703565.10                                                    Warning    Reading         |W| ImportSeriesNumber_CRmCTmREGmTPmTWmMRmPTmSCmUSm: Type 2 Attribute 'SeriesNumber' (0020,0011) is missing in DICOM stream
11/27/2016 22:00:37  2.16.1.113662.2.12.0.3057.1241703565.10                                                    Warning    Reading         |W| ImportManufacturer_CRmCTmRDmRImRSmREGmDREGmTPmTWmMRmPTmSCuRPmRNmRTmRXmUSm: Type 2 Attribute 'Manufacturer' (0008,0070) is missing in DICOM stream
11/27/2016 22:00:37  2.16.1.113662.2.12.0.3057.1241703565.10                                                    Warning    Reading         |W| ImportKVP_CTm            : Type 2 Attribute 'KVP' (0018,0060) is missing in DICOM stream
11/27/2016 22:00:37  2.16.1.113662.2.12.0.3057.1241703565.10                                                    Warning    Reading         |W| ImportAcquisitionNumber_CTm: Type 2 Attribute 'AcquisitionNumber' (0020,0012) is missing in DICOM stream
11/27/2016 22:00:37  2.16.1.113662.2.12.0.3057.1241703565.10                                                    Info       Reading         Import                   : DICOM conversion succeeded
11/27/2016 22:00:37  General                                                                                    Info       GENERAL         ReadFile                 : Reading DICOM Media File 
```
(repeated for each slice)

### LINKED ISSUE #238: dicom2trip.py still does not create .vdx files
silly mistake in api

nbassler:
Closes #191 
Closes #193 
Closes #238


pytrip/ctx.py
```diff
@@ -70,39 +72,49 @@ def create_dicom(self):
...
+                                        "%.3f" % (self.slice_pos[i])]
+
+            _ds.SOPInstanceUID = '2.16.1.113662.2.12.0.3057.1241703565.' + str(i + 1)
+
+            # .HED files do not carry any time stamp (other than the usual file meta data)
+            # so let's just fill it with current times. (Can be overriden by user)
+            _ds.SeriesDate = datetime.datetime.today().strftime('%Y%m%d')
+            _ds.ContentDate = datetime.datetime.today().strftime('%Y%m%d')
+            _ds.SeriesTime = datetime.datetime.today().strftime('%H%M%S')
+            _ds.ContentTime = datetime.datetime.today().strftime('%H%M%S')
```
grzanka: can't these four lines be moved outside the loop ?
it may happen that some slices get later date than others as time is calculated inside the loop

nbassler: yes, but then it was also generated at different time?

Maybe just SeriesDate should be moved outside

grzanka: I think it is safer to assign the same generation time and date to all slices.

In 99% of cases they will anyway have the same time as the execution time of one loop is probably ~ms.

But in some cases it may happen that second is incremented in the clock and part of slices will get a timestamp i.e. 07:15:15 and rest of slices 07:15:16

pytrip/cube.py
```diff
@@ -462,16 +463,23 @@ def create_dicom_base(self):
...
         ds.PhotometricInterpretation = 'MONOCHROME2'
         ds.SamplesPerPixel = 1
         ds.ImageOrientationPatient = ['1', '0', '0', '0', '1', '0']
         ds.Rows = self.dimx
         ds.Columns = self.dimy
         ds.SliceThickness = str(self.slice_distance)
-
         ds.PixelSpacing = [self.pixel_size, self.pixel_size]
+
+        # Add eclipse friendly IDs
```
grzanka: To continue the code style from above, this can be also written as:
```
ds.StudyID = ''
ds.ReferringPhysiciansName = '' 
ds.PositionReferenceIndicator = ''
ds.SeriesNumber = ''
```

nbassler: No, that does not work, I tried.
These attributes do not exist by default in the pydicom object. Creating the attributes, does not tell pydicom to save it when saving the dicom file (also the Tag keys are missing).

Therefore using the add_new() method. Then it works. :)

grzanka: Very interesting ! Can you reproduce what I have below ?
It would be good to know if these problems are related to pydicom version.
If yes, then we should stick to working version in `requirements.txt` file in pytrip

```
→ pip freeze | grep pydicom                                                   ~
pydicom==0.9.9

# grzanka at nz61-6 in ~ [20:07:45]
→ python                                                                      ~
Python 3.6.0 (default, Jan 16 2017, 12:12:55) 
[GCC 6.3.1 20170109] on linux
>>> from dicom.dataset import FileDataset, Tag
>>> ds1 = FileDataset("file", {}, preamble=b"\0" * 128)
>>> ds1.ReferringPhysiciansName = ''
>>> ds1
(0008, 0090) Referring Physician's Name          PN: ''
>>> ds2 = FileDataset("file", {}, preamble=b"\0" * 128)
>>> ds2.add_new(Tag(0x0008,0x0090), 'PN', '')
>>> ds2
(0008, 0090) Referring Physician's Name          PN: ''
```

nbassler: Ok, I only tried with StudyID and I saw it failed. But maybe I missed the case or something.

Also, I was tired of guessing how vendors spell the tags. See the pinnacle case #224 

ROIContours vs ROIContourSequence   (<- no plural s)
RTROIObservations vs RTROIObservationsSequence  (<- including plural s, wtf.)

Feel free to fix what you want to fix, I cannot touch this branch anymore before next week.

grzanka: Thanks for hint. I will take care

grzanka:
To make a test, use this code: https://raw.githubusercontent.com/pytrip/pytrip/cf269d2835b0ebb61c4888b5948dd12d1356e87e/test.py

PR #240, opened 2017-02-09
Fix #237, raystation not loading pytrip generated dicom files.

### LINKED ISSUE #237: RayStation cannot load PyTRiP generated dicoms
Laconic error message "No conversion for this enum type."
No diagnostic files outputted.

nbassler:
Closes #237

### DIFF
--- a/pytrip/ctx.py
+++ b/pytrip/ctx.py
@@ -102,7 +102,7 @@ class CtxCube(Cube):
 
         ds.KVP = ''  # KVP tag 0x0018, 0x0060
 
-        ds.AcquisitionNumber = 1  # AcquisitionNumber tag 0x0020, 0x0012
+        ds.AcquisitionNumber = '1'  # AcquisitionNumber tag 0x0020, 0x0012 (type IS - Integer String) 
 
         for i in range(len(self.cube)):
             _ds = copy.deepcopy(ds)
--- a/pytrip/ctx.py
+++ b/pytrip/ctx.py
@@ -102,7 +102,7 @@ class CtxCube(Cube):
 
         ds.KVP = ''  # KVP tag 0x0018, 0x0060
 
-        ds.AcquisitionNumber = '1'  # AcquisitionNumber tag 0x0020, 0x0012 (type IS - Integer String) 
+        ds.AcquisitionNumber = '1'  # AcquisitionNumber tag 0x0020, 0x0012 (type IS - Integer String)
 
         for i in range(len(self.cube)):
             _ds = copy.deepcopy(ds)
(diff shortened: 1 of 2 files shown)


nbassler:
I also changed ```SeriesNumber``` in cube.py and ```AcquisitionNumber``` in ctx.py from str to int. This was not a problem for loading, but should be more dicom conform.

nbassler:
BTW: RayStation can load multiple CTs. I did the conversion twice, so it got two UUIDs, and both load fine in the same instance of RayStation, even if the PatientID is the same. (Think eclipse...)

pytrip/ctx.py
```diff
@@ -102,7 +102,7 @@ def create_dicom(self):
 
         ds.KVP = ''  # KVP tag 0x0018, 0x0060
 
-        ds.AcquisitionNumber = ''  # AcquisitionNumber tag 0x0020, 0x0012
+        ds.AcquisitionNumber = 1  # AcquisitionNumber tag 0x0020, 0x0012
```
grzanka: I've checked type of `ds.AcquisitionNumber` here https://www.dabsoft.ch/dicom/6/6/#(0020,0012)

```(0020,0012)	Acquisition Number	AcquisitionNumber	IS	1```

It is IS which means IntegerString (following ftp://dicom.nema.org/medical/DICOM/2013/output/chtml/part05/sect_6.2.html):

```
A string of characters representing an Integer in base-10 (decimal), shall contain only the characters 0 - 9, with an optional leading "+" or "-". It may be padded with leading and/or trailing spaces. Embedded spaces are not allowed.
The integer, n, represented shall be in the range:
-231<= n <= (231-1).
```

I think is it much safer to use: `ds.AcquisitionNumber = '1'` instead of `ds.AcquisitionNumber = 1`

File: pytrip/cube.py [hunk 2]
```diff
@@ -478,7 +478,7 @@ def create_dicom_base(self):
         ds.StudyID = ''  # Study ID tag 0x0020,0x0010
         ds.ReferringPhysiciansName = ''  # Referring Physician's Name tag 0x0008,0x0090
         ds.PositionReferenceIndicator = ''  # Position Reference Indicator tag 0x0020,0x1040
-        ds.SeriesNumber = ''  # SeriesNumber tag 0x0020,0x0011
+        ds.SeriesNumber = 1  # SeriesNumber tag 0x0020,0x0011
```
grzanka: SeriesNumber is also `Integer String` (which is a string), not `Signed Long` (which is a number), so I propose to use:
`ds.SeriesNumber = '1'`

to avoid magical conversions done by pydicom.

pytrip/ctx.py
nbassler: Maybe also check if pydicom does any conversion?

because when we inspected here, we saw:
'' -> [ ]
1 -> 1
'1' -> (not known what happens)

nbassler: tested it, '1' is also converted properly. Agree to stay close to the value representations.

PR #258, opened 2017-02-24
Suggested fix for #254, bad patient_ids

### LINKED ISSUE #254: Eclipse recognizes every DICOM set generated by pytrip as the same patient
matchingname
Here Eclipse thinks that the pytrip-generated DICOM set of Alinas patient from #245  is the same as the test files sent by Leszek and Niels  in regard to #193 (which is clearly not the case).  
The patient ID1 is always 123456 for DICOM sets generated with pytrip. Therefore Eclipse thinks it is the same patient. This seems to happen no matter what the other UIDs say because clearly they are listed with different names ("tst002" vs "cube"), being the original names of the respective .hed file.

This issue makes it a bit difficult to joggle with more than one pytrip-generated DICOM set at a time, because you have to delete the existing patient first in order to import the new one.

nbassler:
Closes #254

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+pytrip/_version.py export-subst
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-pytrip/_version.py export-subst
--- a/pytrip/cube.py
+++ b/pytrip/cube.py
@@ -463,8 +463,8 @@ class Cube(object):
         ds.PatientsName = self.patient_name
         if self.patient_id is None:
             ds.PatientID = datetime.datetime.today().strftime('%Y%m%d-%H%M%S')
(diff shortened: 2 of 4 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+pytrip/_version.py export-subst
```
nbassler: @grzanka something not ok in ./versioneer_install

grzanka: I'll fix it.

grzanka:
Patient ID is used both in Cube and in VDX, see: https://github.com/pytrip/pytrip/blob/master/pytrip/vdx.py#L272

grzanka:
@nbassler I did a "push force" on this branch to rebase it and resolve conficts.
Please remove local branch and get fresh remote if you want to add sth.

PR #259, opened 2017-02-24
Enable slice mapping for non-equidistant slices

### LINKED ISSUE #250: Raystation RTStruct - bad slice mapping
It was a file converted from DICOM to TRiP and back. And this is the error from RayStation. (Error number 2)
It was a Pinnacle IMRT DICOM file from the usual test files we use. (PyTRiP 1.3.1)
image

nbassler:
This PR does not resolve the original issue presented in https://github.com/pytrip/pytripgui/issues/58

It should resolve #250 however, since several slices were not mapped to the right z-values.

This PR may require an update of pytripgui.

### DIFF
--- a/pytrip/cube.py
+++ b/pytrip/cube.py
@@ -351,7 +351,7 @@ class Cube(object):
         if self.z_table:
             output_str += "z_table yes\n"
             output_str += "slice_no  position  thickness  gantry_tilt\n"
-            for i, item enumerate(self.slice_pos):
+            for i, item in enumerate(self.slice_pos):
                 output_str += "{:3d}{:16.4f}{:13.4f}{:14.4f}\n".format(i + 1, item,
                                                                        self.slice_distance, 0)  # 0 gantry tilt
         else:
--- a/pytrip/cube.py
+++ b/pytrip/cube.py
@@ -194,9 +194,9 @@ class Cube(object):
         :params int idx: index number of slice (no bound check)
         :returns: position of slice in [mm] including offset
         """
-        # TODO: out of bounds, return nearest neighbour. Update docstring.
-        # return idx * self.slice_distance + self.zoffset
-        return self.slice_pos[idx] + self.zoffset
+        print(idx)
+        # note that self.slice_pos contains an array of positions including any zoffset.
+        return self.slice_pos[int(idx)]
 
     def create_cube_from_equation(self, equation, center, limits, radial=True):
         """ Create Cube from a given equation.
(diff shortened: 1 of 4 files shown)


pytrip/vdx.py
```diff
@@ -1100,7 +1102,14 @@ def create_dicom_contours(self):
             # if CT cube is loaded we extract DICOM representation of the cube (1 dicom per slice)
             # and select DICOM object for current slice based on slice position
             # it is time consuming as for each call of this method we generate full DICOM representation (improve!)
-            candidates = [dcm for dcm in self.cube.create_dicom() if dcm.SliceLocation == self.get_position()]
+
+            from pytrip.util import float_compare
+
+            _eps = 0.00001
+            
+            candidates = [dcm for dcm in self.cube.create_dicom() if float_compare(dcm.SliceLocation,
```
grzanka: You could use this one https://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.isclose.html instead of writing method yourself.

pytrip/cube.py
```diff
@@ -186,44 +186,14 @@ def indices_to_pos(self, indices):
...
-
-    def get_value_at_pos(self, pos):
-        """ Retrieves the value of a voxel at postion [x,y,z] in [mm], including any offsets.
-
-        :param [int*3] idx: list of integers descibing the i,j,k of the data cube.
-        :returns: The voxel value at index[i,j,k]
-        """
-        return self.get_value_at_indice(self.pos_to_indices(pos))
+        # note that self.slice_pos contains an array of positions including any zoffset.
+        return self.slice_pos[int(slice_number) - 1]
```
grzanka: We expect int, so why to cast to int again ?
Avoiding cast to int might serve as a check if `slice_number` is integer.

pytrip/cube.py
```diff
@@ -375,15 +345,17 @@ def write_trip_header(self, path):
...
-        #     output_str += "z_table yes\n"
-        #     output_str += "slice_no  position  thickness  gantry_tilt\n"
-        #     for i in range(len(self.slice_pos)):
-        #         output_str +=
-        # "  %d\t%.4f\t%.4f\t%.4f\n"%(i+1,self.slice_pos[i],
-        # self.slice_distance,0)"""
+        if self.z_table:
+            output_str += "z_table yes\n"
+            output_str += "slice_no  position  thickness  gantry_tilt\n"
+            for i in range(len(self.slice_pos)):
```
grzanka: This can be replaced with enumerate, but it is just a cosmetics

pytrip/vdx.py
```diff
@@ -658,7 +658,7 @@ def get_2d_slice(self, plane, depth):
             points1.extend(reversed(points2))
             points1.append(points1[0])
             s = Slice(cube=self.cube)
-            s.add_contour(Contour(points1))
+            s.add_contour(Contour(points1), cube=self.cube)
```
grzanka: Does `add_contour` method accept additional parameter `cube` ?
Travis detected a problem: https://travis-ci.org/pytrip/pytrip/jobs/205080256

nbassler: I think i placed the parantesis wrong. The cube should be passed to Contour()

pytrip/cube.py
nbassler: Using the cast, the function could also be called using floats such as 1.0. Python itself is rather weakly typed, so I feel it would be more natural that more types could be accepted calling this function. Of course this may obscure errors. 
If we should nazi-type all our functions we will have a lot of work ahead of us..

nbassler:
messed up rebasing it seems

PR #260, opened 2017-02-24
Possible fix for 1.2 VDX pixel size issue

### LINKED ISSUE #245: converting TRiP to DiCOM sets bad pixel size
as discussed in https://github.com/pytrip/pytripgui/issues/58

nbassler:
Closes #245

### DIFF
--- a/pytrip/vdx.py
+++ b/pytrip/vdx.py
@@ -1275,12 +1275,18 @@ class Contour:
         :params xy_line: list of numbers (as characters) representing X and Y coordinates of a contour
         """
 
+        if self.cube is None:
+            _pixel_size = 1.0
+            logger.warning("Reading .vdx in 1.2 format: no CTX cube associated, setting pixel_size to 1.0.")
+        else:
+            _pixel_size = self.cube.pixel_size
+
         # and example of xy_line: 3021 4761 2994 4899 2916 5015
         xy_pairs = [xy_line[i:i + 2] for i in range(0, len(xy_line), 2)]  # make list of pairs
         for x, y in xy_pairs:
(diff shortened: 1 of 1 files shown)


grzanka:
This will solve the problem of reading TRiP files and saving them to DICOM. I've just realised that saving to TRiP format also needs a bit of work.
When reading from .vdx in 1.2 format we do /16 and multiplication by pixel size.
Saving .vdx in 1.2 with reverse operation (*16 and divide by pixel size) is not supported yet.

nbassler:
@grzanka Is there any good reason to support writing to legacy 1.2 vdx format? I think it is obsolete. Reading it should be sufficient.

TokePrintz:
Well, all Alinas files are in this 1.2 format. And those are the ones she needs to convert to DICOM for our joined project. Does "reading" here cover making 1.2 voxelplan files into DICOM?

pytrip/vdx.py
```diff
@@ -1279,7 +1279,9 @@ def read_vdx_old(self, slice_number, xy_line):
         xy_pairs = [xy_line[i:i + 2] for i in range(0, len(xy_line), 2)]  # make list of pairs
         for x, y in xy_pairs:
             # TRiP98 saves X,Y coordinates as integers, to get [mm] they needs to be divided by 16
-            self.contour.append([float(x) / 16.0, float(y) / 16.0, float(slice_number)])
+            self.contour.append([self.cube.pixel_size * float(x) / 16.0,
```
grzanka: Reading vdx without dicom is still supported in many places, but not here.
There should be an if switch to deal with situation when self.cube is None

PR #264, opened 2017-02-27
Fix CtxCube not being passed to VDX when dealing with 1.2 legacy vdx

### LINKED ISSUE #263: CT slices and contours separated with two different patient IDs in 1.3.2
jet02_whytwopatients

I hate to be the party-pooper  but I think there is a bug in the newest version (1.3.2) when voxelplan is generated into DICOM. 

The trip2dicom command gives the following error in the terminal: "WARNING: pytrip.vdx:Reading .vdx in 1.2 format: no CTX cube associated, setting pixel_size to 1.0." The DICOM is although still generated and all looks okay. But when importing the directory with the DICOM in Eclipse, it sees "two patients" (see image). One of them is the CT data, the other is the contour. The contour has the new patient ID that was implemented in 1.3.2, while the CT data has no patient ID (and Eclipse therefore demands you come up with one with import). By first importing the CT data and setting the patient ID to that which the contour has and then importing the contour on top of the CT afterwards, one can manage to get both the CT and the contour imported. BUT since pixel_size was set to 1.0 as the warning in the terminal said, our contours are STILL shifted and placed wrongly, as in version 1.3.1.

Can it be that the new fix for the patient ID is somehow only working for the .vdx and then pytrip doesn't find a matching .ctx and sets a default pixel_size value? 
Also, in a situation like this, maybe pytrip could ask the user for a pixel_size?

nbassler:
Closes #263

Will remove "in progress" once @TokePrintz confirms it works.

### DIFF
--- a/pytrip/vdx.py
+++ b/pytrip/vdx.py
@@ -79,11 +79,13 @@ class VdxCube:
         self.vois = []
         self.cube = cube
         self.version = "1.2"
-        if cube is not None:
+        if self.cube is not None:
             self.patient_id = cube.patient_id
+            logger.debug("VDX class inherited patient_id {}".format(self.patient_id))
         else:
             import datetime
             self.patient_id = datetime.datetime.today().strftime('%Y%m%d-%H%M%S')
+            logger.debug("VDX class creates new patient_id {}".format(self.patient_id))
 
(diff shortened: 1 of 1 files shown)


nbassler:
@TokePrintz confirmed it worked.

PR #271, opened 2017-02-27
remove forbidden 'other' string from RT ROI Interpreted Type

### LINKED ISSUE #257: Eclipse errors and warnings when importing DICOM set, generated from TRiP files format 1.4
error_jet03
I was trying to test different formats of TRiP files (.vdx,.hed,.ctx) being converted into DICOM using pytrip 1.3.1 in relation to #245. 
I had some files in format 1.4 originally converted from DICOM into TRiP format around two years ago using the then-version of pytrip. When I converted these back into DICOM and imported them in Eclipse, I got the above shown errors and warnings and the import failed (there are more warnings not seen in the screen window but they are identical to the last couple of warnings). It is the first time I see these errors. Normally we just get some warnings and the import works.

To recap, the difference between the files tested until now and the warnings/errors they give: 
1) Niels' are generated from TRiP format 2.0. Where they originally DICOM? No warnings, no errors, import works.
2) Alinas files are generated from TRiP format 1.2 from DICOM originally converted into TRiP format using dcm2trip. Some warnings #255, no errors, import works.    
3) In this case files are generated from TRiP format 1.4 from DICOM originally converted into TRiP format using an older 2015 version of pytripgui. We get warnings, errors and the import fails. 
An important note is that the original DICOM Alina uses and the ones I use are from the same place. Only the TRiP format (and the way they were made into this format) are different.

Now it remains to be figured out if this issue comes from the difference in format or if the older version of pytrip had bugs when generating TRiP files?

nbassler:
Closes #257

### DIFF
--- a/pytrip/vdx.py
+++ b/pytrip/vdx.py
@@ -847,7 +847,7 @@ class Voi:
         elif type_id == 1:
             return 'CTV'
         elif type_id == 0:
-            return 'other'
+            return 'ORGAN'
         return ''
 
     def read_dicom(self, info, data):


grzanka:
Following http://dicom.nema.org/medical/dicom/2016d/output/chtml/part03/sect_C.8.8.8.html I suppose ?

nbassler:
yes, i looked it up there.

nbassler:
also: it is in sync with the inverse function, just next to it

grzanka:
Right now the method is not prone to providing values outside the scope.
I'd suggest to make type_id + type_name mapping as a dictionary (LUT) to avoid duplicating code in two methods. Nevertheless this is just a minor improvement

nbassler:
The other warnings mentioned #257 (based on 1.3.1) seem to have been fixed in 1.3.2:
https://github.com/pytrip/pytrip/commit/50be339c2f2db901890cace1b4485f98d26132d2

nbassler:
is a dict a good thing to do when the function is non-injective?

PR #273, opened 2017-02-28
Fix missing PatientID when converting to trip->dicom #266

### LINKED ISSUE #266: RayStation still unhappy dicom->trip->dicom conversion
img_20170227_145109

nbassler:
Regarding #266

### DIFF
--- a/pytrip/cube.py
+++ b/pytrip/cube.py
@@ -431,8 +431,9 @@ class Cube(object):
         meta.TransferSyntaxUID = UID.ImplicitVRLittleEndian  # Implicit VR Little Endian - Default Transfer Syntax
         ds = FileDataset("file", {}, file_meta=meta, preamble=b"\0" * 128)
         ds.PatientsName = self.patient_name
-        if self.patient_id is None:
-            ds.PatientID = datetime.datetime.today().strftime('%Y%m%d-%H%M%S')
+        if self.patient_id is None or self.patient_id == '':
+            self.patient_id = datetime.datetime.today().strftime('%Y%m%d-%H%M%S')
+        ds.PatientID = self.patient_id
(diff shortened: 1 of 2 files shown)


pytrip/cube.py
```diff
@@ -431,8 +431,9 @@ def create_dicom_base(self):
         meta.TransferSyntaxUID = UID.ImplicitVRLittleEndian  # Implicit VR Little Endian - Default Transfer Syntax
         ds = FileDataset("file", {}, file_meta=meta, preamble=b"\0" * 128)
         ds.PatientsName = self.patient_name
-        if self.patient_id is None:
-            ds.PatientID = datetime.datetime.today().strftime('%Y%m%d-%H%M%S')
+        if self.patient_id is None or self.patient_id == '':
+            self.patient_id = datetime.datetime.today().strftime('%Y%m%d-%H%M%S')
+        ds.PatientID = self.patient_id
```
grzanka: This may introduce some bugs in the future, as before the change `create_dicom_base` method wasn't modifying self object and now it is (in case patient_id is missing).

Fixing missing patient_id shouldn't be done here.

nbassler: alternatively one could create self.patient_id in cube.__init__.
It is mandatory for dicoms only, but need to be in sync across trip files. 
Creating it at cube.__init__ level can be overridden still, e.g. when importing from dicom etc.

So, either method is OK with me.

grzanka: If you want I can code it in `cube.__init__`

nbassler: go ahead, i am a bit out of time

grzanka: some improvements in a8e3d7d

PR #274, opened 2017-02-28
fix RTSS import issue

nbassler:
Fixes https://github.com/pytrip/pytripgui/issues/58 together with https://github.com/pytrip/pytripgui/pull/79

### DIFF
--- a/pytrip/ctx.py
+++ b/pytrip/ctx.py
@@ -106,9 +106,10 @@ class CtxCube(Cube):
 
         for i in range(len(self.cube)):
             _ds = copy.deepcopy(ds)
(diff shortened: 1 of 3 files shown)


nbassler:
@grzanka in this stage, all contour+offset issues are fixed, after scrutinizing how TRiP works.

The recent changes will resolve https://github.com/pytrip/pytripgui/issues/78 on the pytrip side.
Therefore I think this should be merged, even if it still does not resolve #266 yet.

nbassler:
@grzanka feel free to take over, I cannot do much on this before tmrw afternoon, but would be great to have it merged.

nbassler:
Note that the exact contour coordinates also are retained converting dicom->trip->dicom. Could be useful to add this to tests.

PR #299, opened 2017-03-04
removed created_by and partially creation_info as TRiP98 is not diges…

### LINKED ISSUE #295: TRiP does not like created_by and creation_info
which is strange, since TST000 is an official plan which has been used earlier with TRiP. Maybe something has changed over the version.

```
TRiP98> bassler@baslup:~/Projects/dasu01$ TRiP98 
--- This is TRiP98-64, Version 1310(compiled May 27 2014 19:45:59), (C) M.Kraemer@gsi.de ---
    @ baslup(Linux #1 SMP Debian 4.9.6-3 (2017-01-28) 4.9.0-1-amd64).
TRiP98> ct "../shieldhit/res/TST000/TST000000" /read

read_hed_file: 
 error = -28 
error in read_hed_file: ../shieldhit/res/TST000/TST000000.hed illegal file structure! 
1999 -28
<E> trpcube.c(Read): read hed file failed
<E> trpct.c(Read): cube read failed
<E> trpct.c(4DRead): cube read failed
<E> trpcmdct.c: cube read failed
```

removing the two tags from .hed will make the file load.

grzanka:
Fixes #295

### DIFF
--- a/pytrip/cube.py
+++ b/pytrip/cube.py
@@ -50,6 +50,7 @@ class Cube(object):
             self.header_set = cube.header_set
             self.version = cube.version
             self.modality = cube.modality
+            self.created_by = cube.created_by
             self.creation_info = cube.creation_info
             self.primary_view = cube.primary_view
             self.data_type = cube.data_type
@@ -82,11 +83,13 @@ class Cube(object):
             self.cube = np.zeros((self.dimz, self.dimy, self.dimx), dtype=cube.pydata_type)
 
         else:
+            import getpass
             from pytrip import __version__ as _ptversion
 
             self.header_set = False
             self.version = "2.0"
             self.modality = "CT"
+            self.created_by = getpass.getuser()
             self.creation_info = "Created with PyTRiP98 {:s}".format(_ptversion)
             self.primary_view = "transversal"  # e.g. transversal
             self.data_type = ""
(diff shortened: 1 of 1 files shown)


grzanka:
Tests failed due to some unneeded import, fixed now.

But I agree that before merging careful tests should be done.

nbassler:
Ok, the original shieldhit/res/TST000 was derived from, had version 1.4 in the header. (See apropos also the very long creation_info text string in that file)

Changing the version string from 1.2 to 1.4 in .hed will make TRiP accept the files with ```creation_info``` and ```created_by```. I think this is a better solution. (Where did that 1.2 come from anyway?)

(Alternatively users can optionally choose a version format.)

nbassler:
I think the issue came up because dcm2trip coverts to 1.2 format. 
When reading these files and then writing again, the extra tags are added, however the version string is not bumped to 1.4.

I think all files trip writes should be version 1.4 (unless user explicitly asks for something else, may be a new low-priority issue)

grzanka:
@nbassler these fields are kept now and saved only if version >= 1.4

nbassler:
could you also make sure, that version is set to 1.4 prior write, even if file read was 1.2?

grzanka:
Nope, file version is kept. But I think this is a separate issue.

Issue #309, opened 2017-03-06
New class: OERCube - oxygen enhancement ratio


nbassler:
relative, i.e. OER from 3 to 1 (= no oxygen effect)

nbassler:
closed by #316

Issue #308, opened 2017-03-06
New class: SFCube (Surviving fraction cube)


nbassler:
to contain relative survival of cells. 1.0 = 100 %

grzanka:
http://bio.gsi.de/DOCS/TRiP98/PRO/DOCS/trip98cmddose.html

> write
> Writes the current dose cube(s) as specified by file. Naming convention:
> 
> contains the extension .DOS or .dos:
> for the physical dose. For the biological dose: .DOS and .HED, where is "bio", "rbe" and "svv".
> otherwise:
> .dos and .hed, where is "phys", "bio", "rbe" or "svv".

Would this be compatible with SVV TRiP98 cubes ?

I see that new TRiP98 can extract them:
http://bio.gsi.de/DOCS/TRiP98/PRO/DOCS/trip98cmddose.html#svv

nbassler:
I think it would be a good thing to keep it compatible, at least by read() and write() methods.
I have no experience with the svv produced by TRiP though.

Issue #307, opened 2017-03-06
New class: RBECube


nbassler:
to hold actual RBE values. Needed by models to be implemented later.

grzanka:
http://bio.gsi.de/DOCS/TRiP98/PRO/DOCS/trip98cmddose.html

> write
> Writes the current dose cube(s) as specified by file. Naming convention:
> 
> contains the extension .DOS or .dos:
> for the physical dose. For the biological dose: .DOS and .HED, where is "bio", "rbe" and "svv".
> otherwise:
> .dos and .hed, where is "phys", "bio", "rbe" or "svv".

Would this be compatible with RBE TRiP98 cubes ?

I see that new TRiP98 can extract them:
http://bio.gsi.de/DOCS/TRiP98/PRO/DOCS/trip98cmddose.html#rbe

nbassler:
I guess so, yes. Compability could be ensured by read() and write() methods.

This also defines the suffix issue then. (```.rbe.dos```, I guess)

nbassler:
Closed via #316

Issue #303, opened 2017-03-05
bad LET mapping


nbassler:
Figure produced with cubeslice. Example handed to @grzanka (22 MB, not sure where to store it in github) produces pictures like this:

LETCube 1:
![cube_012_htv_ramp_050](https://cloud.githubusercontent.com/assets/7140398/23589974/f7d02eb0-01d7-11e7-879a-8a0424cec66b.png)

adding LETCube 2
![cube_012_sub_2_ramp_050](https://cloud.githubusercontent.com/assets/7140398/23589983/199ae5b2-01d8-11e7-995b-1f5d0e792af6.png)

result:  

![sumlet_050](https://cloud.githubusercontent.com/assets/7140398/23589987/26c5e552-01d8-11e7-88d3-d9b3bcd96a09.png)

two things are wrong here:
1) 0 LET should be mapped to a transparent colour
2) there should not be any unmapped colours in the result slice.

This is produced using 
```l = (l1 * d1 + l2 * d2) / (d1 +d2)```
dose cubes d1 and d2 look ok, and are not attached here.

**Comment by grzanka** [comment]
Tracing it down:

```
import pytrip as pt


d1 = pt.DosCube()
d2 = pt.DosCube()

l1 = pt.LETCube()
l2 = pt.LETCube()

d1.read("cube_012_sub_2_ramp.phys.dos")
d2.read("cube_012_htv_ramp.phys.dos")

l1.read("cube_012_sub_2_ramp.dosemlet.dos")
l2.read("cube_012_htv_ramp.dosemlet.dos")

print("l1 min", l1.cube.min())
print("l2 min", l2.cube.min())

print("d1 min", d1.cube.min())
print("d2 min", d2.cube.min())

print("d2*l2 min", (d2*l2).cube.min())
```

gives:

```
l1 min 0.0
l2 min 0
d1 min 0
d2 min 0
d2*l2 min -32756
```

**Comment by nbassler** [comment]
he, that really looks weird!

**Comment by grzanka** [comment]
d2 has dtype int16, while l2 has dtype float32 -> this might be an issue. Investigating

**Comment by nbassler** [comment]
Guess all cubes should be converted to floats upon read. How they are stored on disk should not affect the data type... Users may get confused else.

**Comment by grzanka** [comment]
Sorry, both cubes: d2 and l2 are int16 and what we see is a typical integer overflow.

See:

```
print("l2 min", l2.cube.min())
print("d2 min", d2.cube.min())
[...]

PR #316, opened 2017-03-13
Suggestion to expose general cube class for all non-TRiP data objects…

### LINKED ISSUE #310: New class: CellCube
to hold cells. 

each voxel holds then a tuple, containing:
- number of cells in voxel
- alpha_photons
- beta_photons
- pO2 status [mmHg]

however this is very memory intensive. Surely this can be made smarter..

nbassler:
Closes #310, #309, #308, #307

Requires: all data are always stored as float32, see this comment: https://github.com/pytrip/pytrip/issues/303#issuecomment-285979280

### DIFF
--- a/pytrip/__init__.py
+++ b/pytrip/__init__.py
@@ -33,7 +33,6 @@ from pytrip.raster import Rst
 from pytrip.let import LETCube
 from pytrip.ctimage import CTImages
 from pytrip.dicomhelper import read_dicom_folder
-from pytrip.util import load
 
 # from https://docs.python.org/3/tutorial/modules.html
 # if a package's __init__.py code defines a list named __all__,
--- a/pytrip/__init__.py
+++ b/pytrip/__init__.py
@@ -24,6 +24,7 @@ TODO: documentation here.
 # flake8: noqa
 
 import logging
+from pytrip.cube import Cube
 from pytrip.ctx import CtxCube
 from pytrip.dos import DosCube
 from pytrip.vdx import VdxCube, Voi
@@ -43,3 +44,7 @@ __all__ = ['CtxCube', 'VdxCube', 'Voi', 'DosCube', 'DensityCube', 'LETCube', 'di
 # to prevent it, we add null logging handler, as suggested by Python documentation:
 # as described here: https://docs.python.org/3/howto/logging.html#library-config
 logging.getLogger(__name__).addHandler(logging.NullHandler())
+
+from ._version import get_versions
+__version__ = get_versions()['version']
+del get_versions
--- a/pytrip/__init__.py
+++ b/pytrip/__init__.py
@@ -33,6 +33,7 @@ from pytrip.raster import Rst
 from pytrip.let import LETCube
 from pytrip.ctimage import CTImages
 from pytrip.dicomhelper import read_dicom_folder
+from pytrip.util import load
 
 # from https://docs.python.org/3/tutorial/modules.html
 # if a package's __init__.py code defines a list named __all__,
@@ -44,7 +45,3 @@ __all__ = ['CtxCube', 'VdxCube', 'Voi', 'DosCube', 'DensityCube', 'LETCube', 'di
 # to prevent it, we add null logging handler, as suggested by Python documentation:
 # as described here: https://docs.python.org/3/howto/logging.html#library-config
 logging.getLogger(__name__).addHandler(logging.NullHandler())
-
-from ._version import get_versions
-__version__ = get_versions()['version']
-del get_versions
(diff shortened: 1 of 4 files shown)


pytrip/__init__.py
```diff
@@ -24,6 +24,7 @@
 # flake8: noqa
 
 import logging
+from pytrip.cube import Cube
```
grzanka: Will it really solve all problems ?

Cube itself doesn't have a method to write itself to disk.
`write` method is implemented only in subclasses.

It can be extended to the `Cube` class, but then what kind of file extension should we use ?

pytrip/__init__.py
```diff
@@ -43,3 +44,7 @@
 # to prevent it, we add null logging handler, as suggested by Python documentation:
 # as described here: https://docs.python.org/3/howto/logging.html#library-config
 logging.getLogger(__name__).addHandler(logging.NullHandler())
+
```
grzanka: Please use IDE or remove 'versioneer' stuff.

pytrip/__init__.py
nbassler: Yes, I thought about that. Since there is no TRiP handler for these files anyway (perhaps except for SVV, but that could be added later anyway), I think the best would be to pickle it for saving.

nbassler: this means, user may decide fully alone for filename (maybe except for the SVV later, if it becomes relevant and we really want to add it)

pytrip/__init__.py
nbassler: emacs is my IDE...

pytrip/__init__.py
grzanka: Sure, user can decide but will need to use numpy method:

With subclassed he could do:
```c = CtxCube(...)
c.write('blabla')
```

With master class either:
```
c = Cube(...)
c.write_trip_header('blabla.some_ext1')
c.write_trip_data('blabla.some_ext2')
```

or pickle or some direct method.

Reading from trip format (`read_trip_data_file` method in Cube class) also will not work, as it needs to figure out subclass based on extension.

nbassler: No trip format is needed for writing pO2, sf (unless svv, but this is really not needed now), OER etc. 

So, a simple pickle read/write method could be added to master class, if this really should be a part of the class.

```
c = Cube(...)
c.save()
c.load()
```

nbassler: issue #308 could be left open for handling the trip .SVV cubes, but this is not important now.

grzanka:
Appveyor build fails, because of scipy issue:

pypi has new release of scipy (0.19) https://pypi.python.org/pypi/scipy, but it is for Linux only
anaconda has still old one (0.18.1) -> https://anaconda.org/anaconda/scipy (Windows version)

pytrip/__init__.py
grzanka: I like the idea of pickle - it will be bloody fast.
But be carefull with portability (saving in python2, loading in python3).

What can be considered is numpy save method: https://docs.scipy.org/doc/numpy/reference/generated/numpy.save.html

But it will only save cube data, not any other attribute.

nbassler: Looks good, numpy can also handle objects, not just arrays:
```
>>> import pytrip as pt
>>> import numpy as np
>>> c = pt.CtxCube()
>>> c.read("Projects/shieldhit/res/TST000/TST000000")
>>> np.save("foobar",c)
>>> d = np.load("foobar.npy")
>>> d
array(<pytrip.ctx.CtxCube object at 0x7f6ca2c76fd0>, dtype=object)
>>> 
```

nbassler: numpy did not store the attributes.

pytrip/cube.py
```diff
@@ -918,3 +918,13 @@ def write_trip_data(self, path):
...
+
+    @classmethod
+    def save(cls, path):
+        """ Saves the object as a pickle object.
+        :params: path where file is stored.
+        """
+
+        import pickle
+        with open(path, 'wb') as f:
+            pickle.dump(cls, f, pickle.HIGHEST_PROTOCOL)
```
grzanka: Is it compatible between python 2 and 3 ?
If you save in python3 and load in python2, will it work ?

Python 3.4 seems to introduce protocol versions which are not available in Python2:
https://docs.python.org/3/library/pickle.html#data-stream-format

pytrip/__init__.py
grzanka: What about using numpy.save to save numpy cube (portable !) and some JSON serialiser to rest of class attributes (header-equivalent) ? This would probably mean two files :/

pytrip/cube.py
nbassler: Compability is not essential now. For me it is sufficient to add that as a future feature.

pytrip/__init__.py
nbassler: Indeed, two files would be shit. No more clutter.

grzanka:
@nbassler can I take over and fix Appveyor problems ?

nbassler:
@grzanka yes, thanks. I am done here.

PR #317, opened 2017-03-13
add biology models

### LINKED ISSUE #311: add RCR model for producing data for the new classes

### LINKED ISSUE #315: add Carabe RBE model

nbassler:
Closes #311  (adding RCR)
Closes #315  (adds proton stuff such as Carabe)

### DIFF
--- a/pytrip/models/aux.py
+++ b/pytrip/models/aux.py
@@ -1,0 +1,40 @@
+#
+#    Copyright (C) 2010-2016 PyTRiP98 Developers.
+#
+#    This file is part of PyTRiP98.
+#
+#    PyTRiP98 is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, either version 3 of the License, or
+#    (at your option) any later version.
+#
+#    PyTRiP98 is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with PyTRiP98.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Auxilliary tools
+"""
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def rbe_from_sf(sf_ion, dose_ion, ax, bx):
+    """
+    Returns the RBE for given ion survivng fraction and (alpha/beta)x-ray
+    :params dose: ion physical dose in [Gy]
+    :params sf_ion: surviving fraction in ion beam
+    :params ax: alpha for X-rays in [Gy^-1]
+    :params bx: beta for X-rays in [Gy^-2]
+    """
+
+    # Solve for D_x: bD2 + aD + ln(sf_ion) = 0
+    # RBE = D_x / D_ion
+    logger.warning("rcr_rbe not implemented yet.")
+    pass  # TODO: not implemented yet.
--- a/pytrip/models/aux.py
+++ b/pytrip/models/aux.py
@@ -19,22 +19,38 @@
 """
 Auxilliary tools
 """
-
+import numpy as np
 import logging
 
 logger = logging.getLogger(__name__)
 
 
-def rbe_from_sf(sf_ion, dose_ion, ax, bx):
+def rbe_from_sf(sf_ion, dose_ion, alpha_x, beta_x):
     """
     Returns the RBE for given ion survivng fraction and (alpha/beta)x-ray
-    :params dose: ion physical dose in [Gy]
-    :params sf_ion: surviving fraction in ion beam
-    :params ax: alpha for X-rays in [Gy^-1]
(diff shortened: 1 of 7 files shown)


grzanka:
Please rebase (checkout master, pull first) - it will solve Appveyor problems

pytrip/models/rcr.py
```diff
@@ -0,0 +1,108 @@
...
+    return sf
+
+
+def _f(let):
+    """
+    f function from Dasu paper, takes let-cube as parameter
+    Equation (7) in https://doi.org/10.1093/jrr/rru020
+    """
+
+    let[let == 0] = 0.01
```
grzanka: My previous comment was lost :/

I recommend not to modify let variable which is passed a a parameter, but do the masking on the result, like:

```
>>> b = np.asarray( [1.0, 1.0] )
>>> a = np.asarray( [1.0, 0.0] )
>>> b/a
array([  1.,  inf])
>>> c = b/a
>>> c
array([  1.,  inf])
>>> c[c==np.inf] = 100
>>> c
array([   1.,  100.])
```

This would require:
```
result = (1 - np.exp(-let/ld) * (1 + let/ld)) * ld/let
result[result == np.inf] = 100.0
return result
```

nbassler:
questions @grzanka 
- how to properly handle masked arrays (see ```tcp_voi()``` in aux.py )
- naming api: would like to double check. E.g. ```tcp_from_voi()``` vs. ```tcp_voi()```
- should any of this be exposed, or would it be possible to ```from pytrip.models import *``` or similar?

pytrip/models/rcr.py
nbassler: that was a residue from original code, it may be changed.

pytrip/models/aux.py
```diff
@@ -0,0 +1,40 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def rbe_from_sf(sf_ion, dose_ion, ax, bx):
+    """
+    Returns the RBE for given ion survivng fraction and (alpha/beta)x-ray
+    :params dose: ion physical dose in [Gy]
+    :params sf_ion: surviving fraction in ion beam
+    :params ax: alpha for X-rays in [Gy^-1]
```
grzanka: I'd suggest `alpha_x` instead of `ax` as it is more pythonic.

grzanka:
@nbassler 

-  masked arrays -> what about array of True & False values ?
-  tcp_voi sounds OK
-  before exposing it I'd wait until final implementation. Even not exposed code can be used by simple import. Once the code is ready it would be good to cross-check if any refactoring of structure, function grouping etc is needed.

nbassler:
@grzanka btw, I rebased, but AppVeyor is still unhappy.

grzanka:
Indeed, Appveyor gives an interesting error. I suggest to ignore it now - I can investigate later

grzanka:
It is interesting - all these models can operate on numpy and be independent on TRiP98 - is it like that ?

If yes, then in principle one could get the source code from github and work with them without installing pydicom or other dependencies of pytrip.

pytrip/models/rcr.py
```diff
@@ -0,0 +1,147 @@
...
+    :params let: LET in [keV/um]
+
+    :returns: result of the f function
+    """
+
+    ld = 86.0
+    result = (1 - np.exp(-let/ld) * (1 + let/ld)) * ld/let
+
+    # map any zero LET areas to 0.0
+    if hasattr(result, "__len__"):  # for numpy arrays
```
grzanka: I know it is pythonic, but numpy has now a method to check that: `np.isscalar`

grzanka:
@nbassler travis tests are failing. Probably you are missing (empty) `__init__.py` file.
See https://github.com/pytrip/pytrip/blob/master/pytrip/utils/__init__.py
Each module should have such file in it.

nbassler:
@grzanka why is appveyor failing?

grzanka:
I have no clue about appveyor - probably sth strange and related to GIT (rebase) ?

nbassler:
most models are trip independent, but the tcp model may take a VOI optionally, and then needs pytrip.

pytrip/models/proton.py
```diff
@@ -0,0 +1,127 @@
...
+#
+#    You should have received a copy of the GNU General Public License
+#    along with PyTRiP98.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Collection of proton RBE models.
+
+[1] A. Carabe, M. Moteabbed, N. Depauw, J. Schuemann, and H. Paganetti,
+"Range uncertainty in proton therapy due to variable biological effectiveness,"
+Phys. Med. Biol. 57(5), 1159–1172 (2012).
```
grzanka: Travis fails to build on python 2.7, I suspect a problem with `-` character

```
E     File "/home/travis/build/pytrip/pytrip/pytrip/models/proton.py", line 24
E   SyntaxError: Non-ASCII character '\xe2' in file /home/travis/build/pytrip/pytrip/pytrip/models/proton.py on line 25, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

grzanka: Check also other comments for non-ASCII characters

grzanka:
@nbassler I've already applied a fix for non-ASCII stuff

PR #339, opened 2017-03-27
new DDD implementation

### LINKED ISSUE #338: proper implementation of  DDD file reader
Current implementation in looks messy and broken

grzanka:
Fixes #338

### DIFF
--- a/examples/example03_ddd.py
+++ b/examples/example03_ddd.py
@@ -1,0 +1,30 @@
+#
+#    Copyright (C) 2010-2017 PyTRiP98 Developers.
+#
+#    This file is part of PyTRiP98.
+#
+#    PyTRiP98 is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, either version 3 of the License, or
+#    (at your option) any later version.
+#
+#    PyTRiP98 is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with PyTRiP98.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+This example demonstrates how to deal with DDD files.
+"""
+
+import pytrip as pt
+
+ddd_files = "/home/user/trip98/*.ddd"
+
+ddd = pt.DDD()
+ddd.read(ddd_files)
+
+energies = ddd.energies()
--- a/pytrip/__init__.py
+++ b/pytrip/__init__.py
@@ -36,7 +36,7 @@ from pytrip.dicomhelper import read_dicom_dir
 # from https://docs.python.org/3/tutorial/modules.html
 # if a package's __init__.py code defines a list named __all__,
 # it is taken to be the list of module names that should be imported when from package import * is encountered.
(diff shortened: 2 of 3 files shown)


grzanka:
@Tibazz how do you like such implementation of DDD class ?

I think it is faster to rewrite it from scratch than fixing the old one.

I still need to improve it but with what you have here you can easily  read/write single DDDs and collections of them (folders) and manipulate on their data.

What is still missing is the interpolation between energy steps - I'll fix it in the free time.

grzanka:
@nbassler any opinion on that ?

nbassler:
@grzanka @pititomy @Tibazz @okejdokej 
The purpose of loading DDD is to provide kernel files for future optimizer (OptiRT).

- DDD() object should be able to load one or more .ddd files of one or more energies
- I would expect a DDD() class to be able to return a dose for any energy in cartesian grid dose =
 f(energy,x,y,z) or in cylindrical coordinates dose = f(energy,r,z) based on input set. This involves interpolation (or even some crude extrapolation).
-- thus two methods: one for returning an answer for a single point (conveience, also good for inspection) and another for returning a 3D numpy.array of the entire dose cube. 
- Manipulating data in DDD() object I guess is secondary importance for now.
- Writing to .ddd files I think is less needed, but one might already reserve the method names for it.

what else?

grzanka:
I'd prefer to stick with OO principles and have single class responsible for doing single thing.
 - DDD represents depth-dose for single energy (one file)
 - DDDCollection represents set of depth-dose curves for many energies (set of files)

Having one class to support both single and many files would lead to a mess.

The second one (DDDCollection) can be renamed, but I'd keep its responsibility.

Returning dose for a given point in space can be easily implemented in DDD (there only interpolation along Z is needed) for given energy:
```
d1 = DDD()
d1.read("bla_e_120.0.ddd")
d = d1.dose(x=2,y=4,z=5)
```

When having DDDCollection user may request dose in space for some intermediate energy:
```
dc = DDDCollection()
dc.read("bla_e_*.ddd")
d = dc.dose(e=110.5, x=2,y=4,z=5)
```
Code above will require additional interpolation on energy. Internally an intermediate DDD will be returned and its method to calculate dose will be called.

nbassler:
in principle ok, but on the other hand, think again the user perspective into it.
If the API gets bloated, the user is confused and will not use it.
In this case, i think the user to mess around with both "DDD" and "DDDCollection" is kind of also breaking the pattern of "Ctx" "Dos" "Vdx" etc classes.
So, we should appeal to the intuition of the user.

In this case, this means, if DDDCollection is going to hold all the interesting functions, I would name that one DDD(). And maybe the primitives behind it _DDD or something else.

Also: how bloated would if one puts it into a single class in this particular case? Even if the theory is correct, I think it should be hold up to the user experience, and accept compromises. The goal is to have a tool which one can do fast science with, and not a tool which is semantically correct, turing complete, etc

We *must* expose the most important methods in a nice and intuitive way. If I need to call multiple classes just do inspect the dose at a x,y,z position in a .ddd file we are doing something wrong.

nbassler:
God objects: good link, this encompasses our discussion. See your God object vs. 
https://en.wikipedia.org/wiki/Spaghetti_code#Ravioli_code which represents my point of view.

:smiley: 

Separating out DDDFile makes sense if we suspect it being used by other parts of code, which I admit is thinkable, however not really exactly the main application (as energies will have to be interpolated for the application in mind).

pititomy:
Coming back on Niels's first comments : "returning a 3D numpy.array of the entire dose cube"

This is actually really needed for Tibor's work, as he is going to superimpose the DDD dose cube on the CT data, with scaling adjustments for the different densities in the voxels.

In that case, working with sparse matrixes would be very beneficial to improve the work load and speed/efficiency of the whole thing.

PR #341, opened 2017-03-28
Repair coronal and saggital contour extraction

### LINKED ISSUE #340: repair indices_to_pos() function
it returns bad values.
https://github.com/pytrip/pytripgui/issues/84

### LINKED ISSUE #342: dicom2trip: sanity check
add a simple check to avoid weird errors when slice_thickness is larger than slice distance.
(had that case with an internal dataset)

nbassler:
issues
- [x] contours in sagittal and coronal are messed up due to missing sorting
- [x] check offsets in TRiP: is this still properly handled?
- [x] how to handle overlapping slices (i.e. thickness = 5 mm and distance 3.5 mm)

closes #342 
closes #340

### DIFF
--- a/pytrip/cube.py
+++ b/pytrip/cube.py
@@ -907,6 +907,11 @@ class Cube(object):
             logger.warning("Only a single slice found. Setting slice_distance to slice_thickness.")
             self.slice_distance = self.slice_thickness
 
+        if self.slice_thickness > self.slice_distance:
+            # TODO: this is probably valid dicom format, however let's print a warning for now
+            # as it may indicate some problem with the input dicom, as it is rather unusual.
+            logger.warning("Overlapping slices found: slice thickness is larger than the slice distance.")
+
         self.set_byteorder()
         self._set_format_str()
         self.header_set = True
--- a/pytrip/cube.py
+++ b/pytrip/cube.py
@@ -203,11 +203,17 @@ class Cube(object):
         The z position is always following the slice positions.
 
         :params [int] indices: tuple or list of integer indices (i,j,k) or [i,j,k]
-        :returns: list of positions,including offsets, as a list of floats [x,y,z]
+        :returns: list of positions, including offsets, as a list of floats [x,y,z]
         """
         pos = [(indices[0] + 0.5) * self.pixel_size + self.xoffset,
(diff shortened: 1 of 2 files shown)


nbassler:
how to handle wider slices: rely on slice positions, not slice thickness, after discussing locally.

pytrip/cube.py
```diff
@@ -394,7 +394,7 @@ def write_trip_header(self, path):
             output_str += "slice_no  position  thickness  gantry_tilt\n"
             for i, item in enumerate(self.slice_pos):
                 output_str += "{:3d}{:16.4f}{:13.4f}{:14.4f}\n".format(i + 1, item,
-                                                                       self.slice_distance, 0)  # 0 gantry tilt
+                                                                       self.slice_thickness, 0)  # 0 gantry tilt
```
nbassler: I made to hed+ctx files with either slice_distance or slice_thickness using the overlapping slice DICOM-CT i have. Made a proton plan on both of them, and TRiP mapped it correctly in both cases. So I prefer passing slice_thickness here as it is more correct, and slice_distance can be derived anyway from the z_position.

pytrip/vdx.py
```diff
@@ -788,6 +789,17 @@ def create_dicom_contour_data(self, i):
...
+    def _sort_slices(self):
+        """ Sorts all slices stored in self for increasing z.
+        This is needed for displaying Saggital and Coronal view.
+        """
+        # slice_in_frame is only given by VDX, and these are also the only frames which need to be sorted
+        # it seems. DICOM apparently have proper structure already. Nonetheless, this function is also
+        # applied to DICOM contours.
+
+        if hasattr(self.slices[0], "slice_in_frame"):
+            self.slices = sorted(self.slices, key=lambda _slice: _slice.slice_in_frame, reverse=True)
```
grzanka: Better to use `sort` method of `list` class which is sorting in place:::

```
self.slices.sort(key=lambda _slice: _slice.slice_in_frame, reverse=True)
```

grzanka:
@nbassler Feel free to squash merge when the tests are done. I may have limited internet connectivity this evening. Tomorrow morning I can make new releases

Issue #180, opened 2016-11-14
Naming: "folder" -> "directory"


nbassler:
cosmetics in the naming convention of subroutines and documentation. 
"Folder" is rather a GUI thing.

nbassler:
can easily be fixed once #349 has merged.

Issue #131, opened 2016-10-17
Make private functions using underscore prefix


nbassler:

nbassler:
@grzanka  very nice, i just saw, that  docstrings of _ prefixed methods `_foobar() ` will result in they are not included in sphinx.

Issue #118, opened 2016-10-11
Avoid using getters and setters


nbassler:
https://www.quantifiedcode.com/app/issue_class/68iQGerP

in particuar tripplan.py

PR #349, opened 2017-04-10
New tripexecuter2

### LINKED ISSUE #114: tripexecuter: make Image argument optional
I cannot see the purpose why the image argument is mandatory when initializing TripExecuter(), it would be nice if this can be omitted / made optional

### LINKED ISSUE #121: tripexecuter should return some result
For instance a path to where the results are stored.

### LINKED ISSUE #278: tripexecutor: Make recasting Voi object obsolete
See e.g. example 02:  https://github.com/pytrip/pytrip/blob/master/examples/example02_trip98.py#L58

```
# next pick a the proper VOI (from the VdxCube object) which we want to plan on
target_voi_temp = my_vdx.get_voi_by_name(my_target_voi)

# for technical reasons the voi must be cast into a new Voi object
# we are working on a cleaner Voi class implementation to avoid it
target_voi = pte.Voi("GTV_VOI", target_voi_temp)
```

### LINKED ISSUE #346: tripexecuter: understands only few ions
pytrip.tripexecuter currently only understands 'H' 'C' 'O' and 'Ne'

We need some proper handling here for e.g. He-3, He-4 and similar.

### LINKED ISSUE #347: tripexecuter: no real ripple-filter support
Currently it can only be specified by loading appropiate DDD/SPC files. However, it may also affect the "scancap" command in TRiP98.

As a user I would expect some attribute in pte.TripPlan() class where this can be specified.

### LINKED ISSUE #348: tripexecuter: remove multionplan feature
following discussion in #346

nbassler:
closes #348 (no multiplans)
closes #346 (more projectiles)
closes #278 (Voi recast not needed)
closes #347 (add ripple filter attribute)
closes #114 (no Image class needed)
closes #121 (get results easily)

follows #118 (getters and setters)
follows #131 (underscore prefix)
follows #180 (get rid of "folder" in API)


nbassler:
@grzanka maybe you can take a first look. It is still far from finished, but the general structure of Field() and Plan() is pretty much in place. See also the very stripped Execute()

I would like to have some feedback on possible directory structure too (right now it nonexisting).
We have
A TRiP98 style `Plan().name`, such as `tst101002`

- a `working_dir`: this is kind of a base dir, default set to "."
- a `temp_dir`, where all stuff needed for calculation will be included in a flat file. This will be created with tmpfilename from working dir. ./trip98_xxxxxxxxxxx/.
- `remote_dir`:  the aforementioned `temp_dir` can also be tarballed and exported to a `remote_dir` such as remote_host:/tmp/. for trip execution
- final results must be stored somewhere, possibly not in the `working dir`, but better in a pristine new directory like `working_dir/planname/<results.*>`

Finally, I look for some term we can call a path attributes ( == a fully quallified directory + filename + suffix) and the same with the suffix removed. Right now the term "path" is used for both, which makes the code rather difficult to read.

like: 
```
field.rasterfile_path = "/home/bassler/test/tst101002.rst"
field.rasterfile_spath = "/home/bassler/test/tst101002"   # as in "stripped paths"
```
if we could agree on the `spath` terminology it would be helpfule?

nbassler:
features addressed here:
- get rid of getters and setters #118 
- get rid of weird casts #278 
- get rid of multion feature #348 
- get rid of "folder" named API #180 
- get rid of Image arguments (use Ctx instead) #114 
- rifi support #347 
- hide weird methods from use using underscore prefix #131 
- simple return of results for user #121

pytrip/tripexecuter2/execute.py
```diff
@@ -0,0 +1,579 @@
...
+        self.ctx = ctx
+        ##self.rbe = rbe
+        self.listeners = []
+        self.trip_bin_path = "TRiP98"  # where TRiP98 is installed, if not accessible in /usr/local/bin or similar
+        self.logfile_stdout = "trip98.stdout"
+        self.logfile_stderr = "trip98.stderr"
+        self.rsakey_local_path = "~/.ssh/id_rsa"
+        self._runtrip = True  # set this to False for a dry run without TRiP98 (for testing purposes)
+        ##self.run_dir = "" 
+        self.remote_dir = "/tmp/."  # remote directory where all will be run
```
grzanka: This is not good. On some machines user may not have access to `/tmp`, but only to some subdirectory.
Safer to use  sth like `mkdtemp`

nbassler: it is just a random string, see discussion above..

grzanka:
Speaking of spath - take look on `os.path.splitext` documentation.

```
Split the pathname path into a pair (root, ext) such that root + ext == path, and ext is empty or begins with a period and contains at most one period. Leading periods on the basename are ignored; splitext('.cshrc') returns ('.cshrc', '').
```

Python is using concept of `root path`, so what about rpath or root_path ?

nbassler:
sounds good, but "*_root_path" is a bit long for some attributes, maybe we can also allow for *_rpath" ? Just that we agree. Then I will put appropriate notes in the documentation.

nbassler:
in summary:
`/home/bassler/foobar/funk.dat`
- `funk.dat` : filename
- `funk` : basename
- `/home/bassler/foobar` : directory `dir`
- `/home/bassler/foobar/.` : directory `dir`
- `./foobar` : relative dir. Could we find some naming convention for `dir` attributes which are _always_ relative? `rdir` ?
- `/home/bassler/foobar/funk.dat` : path
- `/home/bassler/foobar/funk` : root_path or rpath

grzanka:
my proposition:
`/home/bassler/foobar/funk.dat`
- `funk.dat` : filename
- `funk` : basename
- `/home/bassler/foobar` : absolute directory `abs_dir`
- `./foobar` :  relative directory `rel_dir`
- `/home/bassler/foobar/funk.dat` : absolute `abs_path`
- `foobar/funk.dat` : (relative) path `rel_path`
- `/home/bassler/foobar/funk` : root path `root_path`
- `./foobar/funk` : root path `root_path`

Prefixes:
 - 'abs' -> absolute path to file or directory, starting with `/` or `C:\` (see `os.path.abspath`)
 - 'root' -> root part of path (may be absolute or relative, see `os.path.splitext`)
 - 'rel' -> relative path (see `os.path.relpath`)

nbassler:
`self.rasterfile_path` -> `self.rasterfile_root_path`  makes some attribute names long :(

nbassler:
abs_ and rea_ prefix only if it is either. If it can be both, then just leave them out.

Should this go into Contribute.md?

grzanka:
What about annotating function parameters ?
https://www.python.org/dev/peps/pep-0484/

nbassler:
would be very useful, yes.

grzanka:
The only thing to test is if these annotations are well supported in all python versions we use.

nbassler:
No good for pytripgui which only does python2:
```
bassler@baslup:~$ python
Python 2.7.13 (default, Jan 19 2017, 14:48:08) 
[GCC 6.3.0 20170118] on linux2
>>> def greeting(name: str) -> str:
  File "<stdin>", line 1
    def greeting(name: str) -> str:
                     ^
SyntaxError: invalid syntax
>>> 
```

grzanka:
Another option is to use `mypy`-like comments:
https://github.com/python/mypy/issues/719
http://mypy.readthedocs.io/en/latest/python2.html

nbassler:
njaeh, why not leave it on docstring level isntead:
`:params str mypath: path to whatever`

grzanka:
Sure, docstring level makes it clearer to read, but when using `mypy`-like comments you can also enable type checking which can detect many new interesting issues.

pytrip/tripexecuter2/plan.py
```diff
@@ -0,0 +1,293 @@
...
+        self.target_dose = 2.0  # target dose in Gray
+        self.target_dose_percent = 100.0  # target dose in percent
+        self.target_dose_cube = None  # pass a DosCube() here, for incube() optimization.
+
+        self.target_tissue_type = ""  # target tissue type, for biological optimization
+        self.res_tissue_type = ""  # residual tissue types
+        self.incube = False  # enable or disable incube optimizations
+
+    def __str__(self):
+        """ Pretty print all attributes
```
grzanka: This method is way to long. What about exporting its functionality outside to 
 - private method in a module which takes plan as an argument and returns a string
 - helper class with static method which does the same 
?

pytrip/tripexecuter2/plan.py
```diff
@@ -0,0 +1,293 @@
...
+        self.dose_alg = "cl"
+        self.bio_alg = "cl"
+        self.eps = 1e-3
+        self.geps = 1e-4
+
+        # algorithms for dose optimizations
+        self._opt_principles = {"H2Obased": "Very simplified single field optimization",
+                                "CTbased": "Full optimization, multiple fields"}
+
+        self._opt_methods = {"phys": "Physical dose only",
```
grzanka: All this dictionaries could be removed from `__init__` and moved to a class fields as the are constants.

pytrip/tripexecuter2/plan.py
nbassler: sure, I just had to start somewhere. Nothing is constant yet.

pytrip/tripexecuter2/plan.py
nbassler: sure.

docs/contributing.rst
```diff
@@ -179,4 +180,35 @@ To run only a single test type::
...
+4. Functions, class methods, attributes etc which are not supposed to be accessed by users should be prefixed with underscore i.e. ``_foobar``
+5. Directories, paths and filenames should be named following this scheme:
+
+* **Filenames**
+    * ``funk.dat`` : filename
+    * ``funk`` : basename
+* **Directories**
+    * ``/home/bassler/foobar`` : absolute directory ``abs_dir``
+    * ``./foobar`` : relative directory ``rel_dir``
+    * or just ``dir`` if both may be applicable.
```
grzanka: dir is a python builtin function: https://docs.python.org/2/library/functions.html#dir

nbassler: it will be prefixed/suffixed with other stuff, this is just a part of the string

nbassler:
@grzanka  This is my proposal for a basic structure (see triptest.py how to use it)
Can you find some time to run over it very soon? You may also refactor what you want to refactor.

nbassler:
@grzanka remote execution is still disabled, I will do that once you agree with the API.

nbassler:
@grzanka remote execution works now.

nbassler:
@grzanka Basic functionality is now in place. 
Things still to be done, but I think it would be better to put these into new issues? Just not to have this merge too massive.
- [ ] proper handling for different ions (must check how TRiP can handle the projectile mass)
- [ ] implement biological optimization
- [ ] attach resulting .rst fields to Field class after TRiP98 run
- [ ] test and fix incube optimization
- [ ] update example 02 in documentation
- [ ] fix issue with missing $PATH environment variable when running on remote server
- [ ] add a few tests
(You may also refactor, maybe might even fix some of the above issues during refactoring :)

docs/contributing.rst
grzanka: But the documentation doesn't mentione prefixing, so I'd either say `_dir` or `dir_` in order not to suggest user to use `dir` name.

pytrip/cube.py
```diff
@@ -730,6 +730,7 @@ def read(self, path):
 
         :param str path: Path to filename to be read, file extention may be given but is not neccesary.
         """
+        self.basename =  os.path.basename(path).split(".")[0]
```
grzanka: Why is it needed ?

pytrip/tripexecuter2/execute.py
```diff
@@ -0,0 +1,496 @@
...
+This file contains the Execute() class which can execute a Plan() object locally or on a remote server
+whith a complete TRiP98 installtions.
+"""
+import os
+import shutil
+import tarfile
+
+from subprocess import Popen, PIPE
+
+from pytrip.dos import DosCube
```
grzanka: Imports should be reorganized, first uuid and logging, then pytrip

pytrip/tripexecuter2/execute.py
```diff
@@ -0,0 +1,496 @@
...
+    logger.warning("Paramiko package not installed, only local TRiP access possible")
+
+
+class Execute():
+    def __init__(self, ctx, vdx):
+        """ Initialize the Execute class.
+        :params CtxCube() ctx: the CT images as a regular pytrip.CtxCube() object.
+        :params VdxCube() ctx: the contours as a regular pytrip.VdxCube() object.
+
+        If the Ctx and Vdx obejcts are constant, this object can be used to execute multiple
```
grzanka: Python doesn't have the concept of constants, so better to reformulate this comment

pytrip/tripexecuter2/execute.py
```diff
@@ -0,0 +1,496 @@
...
+
+        # once the file list _flist is complete, copy it to the package location
+        for _fn in _flist:
+            logger.debug("Copy {:s} to {:s}".format(_fn, plan._temp_dir))
+            shutil.copy(_fn, plan._temp_dir)
+
+        # Ctx and Vdx files are not copied, but written from the objects passed to Execute() during __init__.
+        # This gives the user better control over what Ctx and Vdx should be based for the planning.
+        self.ctx.write(os.path.join(plan._temp_dir, self.ctx.basename + ".ctx"))  # will also make the hed
+        self.vdx.write(os.path.join(plan._temp_dir, self.vdx.basename + ".vdx"))
```
grzanka: I'd personally prefer linking from temporary dir to original files, instead of copying them. This may potentially use too much space if dealing with many plans and files.

pytrip/tripexecuter2/field.py
```diff
@@ -0,0 +1,126 @@
...
+        out += "|  Contour extension            : {:.2f}\n".format(self.contour_extension)
+        out += "|  Use external .rst file       : {:s}\n".format(str(self.use_raster_file))
+        if self.isocenter:
+            out += "|  Isocenter (x,y,z)            : {:.2f} {:.2f} {:.2f} mm\n".format(self.isocenter[0],
+                                                                                        self.isocenter[1],
+                                                                                        self.isocenter[2])
+        else:
+            out += "|  Isocenter (x,y,z)            : (not set)\n"
+        out += "----------------------------------------------------------------------------\n"
+        return(out)
```
grzanka: brackets not needed

pytrip/cube.py
nbassler: very sad story. In TRiP98 the patientname must be equal to the file name to *some extent*.
using patient_name is not safe, as it may be differen from DICOM import and the like.

Also file name can be suffixed with three digits, and TRiP98 will still accept them as a running number for a plan. So a patient with the patient name TST000 may have the dose cube TST001001 TST001002 TST001003 and the like.

Unfortunately the filename is not store anywhere in the file, so when passing cube objects along to e.g. tripexecuter, this info is lost. Thefore hte basename (i.e. without file extention) must be stored somewhere in the cube object (same for VDX).

pytrip/tripexecuter2/execute.py
nbassler: what rules do we follow?

pytrip/tripexecuter2/execute.py
nbassler: I am not sure how TRiP98 will behave when providing sporadic full paths or relative files (and even worse where they will output them). So in order to avoid any troubles here (and to be more time efficient) I accepted this data duplication (which is on disk and not in mem), and instead have a clean and simple environment where I know what is going on.

pytrip/tripexecuter2/execute.py
grzanka: https://www.python.org/dev/peps/pep-0008/#imports

> Imports should be grouped in the following order:
> 
> 1 standard library imports
> 2 related third party imports
> 3 local application/library specific imports
> You should put a blank line between each group of imports

pytrip/tripexecuter2/execute.py
grzanka: OK, got your point

pytrip/cube.py
grzanka: I see the point. There is already a place where path parsing is done, now it is in `read_trip_data_file`:

```
        # extract header and data file name from path
        header_file_name, data_file_name = self.parse_path(path)
```

I'd propose to move setting basename there (from `read` into `read_trip_data_file`) to keep all path related code in single place.  
Then in the future it could become an element of some more sophisticated path parsing engine.

grzanka:
@nbassler general structure looks fine to me. This executer could easily replace existing one.
I'd recommend to: 
 - refactor it a bit (where needed)
 - remove two new tests files (or move to test directory)
 - remove no 2 from current executer in order to replace the old one
Finally this branch could be merged

pytrip/cube.py
nbassler: I agree.
Note, this may be missing for DICOMs (Ctx and Vdx).

nbassler:
@grzanka agreed.
Regarding tests: I foresee two issues:

1) TRiP98 not available in test environment. All should then be run with _norun option. (Maybe it would be more natural to have this as a command line option to `te.execute(plan, norun=True)` instead of by attribute name.
2) remote execution: how on cluster? Requires some keys to be exchanged and have a hostname.

The two test files can also form the base for an updated documentation (example02)

Can you do the refactoring and the test implementation?
I will gladly fix the documentation.

nbassler:
@grzanka propose not to rename tripexecute2 yet, as long it is not in its final form. This will make it easier for me to adapt pytripgui without breaking the old, and give me more time to add the remaining issues (such as adding the RBE and test the incube optimization)

grzanka:
Do you mean to merge and have two executers?

25.04.2017 15:17 "Niels Bassler" <notifications@github.com> napisał(a):

nbassler:
correct.

nbassler:
Finally...

grzanka:
Script added in b9057aa doesn't work under Windows which makes Appveyor script to fail

nbassler:
would that be 2.0.0?

grzanka:
Yes

nbassler:
I just realized, the the old executer in pytripgui is pretty much broken anyway, so we may just as well get rid of it.

grzanka:
@nbassler It seems that right now a new version of Sphinx (1.6) is being released which is causing many problems. I've restricted Sphinx version in 7df3cd3 to 1.5

grzanka:
Examples are not displayed correctly in Sphinx generated doc: http://pytrip.readthedocs.io/en/feature-348-tripplan/examples.html#example-02-trip-execution

Issue #324, opened 2017-03-22
undefined method in ddd.py?


Tibazz:
https://github.com/pytrip/pytrip/blob/master/pytrip/ddd.py#L36
this method is not defined for this class it seems?

grzanka:
Yes, it is not defined and this is a bug ! Thanks for pointing that !

DDD class is not tested and not used by any other parts of the code, a bit abandoned :(

Do you need it ? If yes, please describe your case, we can fix it or provide some other solution.

Tibazz:
I am working on a master-theisis where i need to use .ddd files inside a custom raycasting algorithm and interpolate them as i go along to achieve, amongst other things, dose distributions and i was looking into PyTrip98 to do those interpolations. So far i have made in-house code for quite a lot of things that frankly already exist inside PyTrip98 mostly because i cant figure out how to install it :) and also because i want to learn stuff by myself. These bugs i've found from copy-pasting the source code.

grzanka:
Great ! We are happy to see new user of Pytrip. 
If you are patient enough to complain using issues on github we can help you to fix most of the missing parts in pytrip and improve the installation procedure.

On the other help you have some inhouse code that could enrich pytrip repo, please let us know (via issue), so we could discuss.

I will fix the problematic in next days. If you could paste here some use-case for DDD class I would check it and improve the code a bit.

grzanka:
See PR #339

Issue #242, opened 2017-02-14
Consider replacing scipy with sth smaller


grzanka:
pytrip depends on scipy package, but needs only tine fraction of it: interpolation module.

putting scipy on requirement list has several consequences:
 - installing scipy on Windows is problematic (no precompiled package that could be installed with pip, need to install via Anaconda)
 - when preparing single executable version of pytripgui, fetching whole scipy is increasing size of compressed .exe by another ~50MB

there are some lightweight libraries that could be used for that purpose, i.e. https://pypi.python.org/pypi/tinyspline but this topic should be studied more carefully.

nbassler:
it is also creating hell when trying to install pytrip98 on debian 8, stable, jessie.

grzanka:
what kind of error do you get ?

grzanka:
and which python ?

nbassler:
it compiles for 20 mins or so.

grzanka:
still it would be useful to reproduce this problem. Can you paste here at least beggining of the console dumpy, starting from pip install command and then next ~100 lines ?

nbassler:
python2 compiles without errors it seems just takes crazy long time. Now waiting for python3 to complete.

grzanka:
Possible solutions is to replace interpolation routines, but these are splines:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.splev.html

We may also check if there is some solid library which does only splines.
https://pypi.python.org/pypi?%3Aaction=search&term=spline&submit=search

nbassler:
... and then it stalls:
```
(... tons of messages...)
    
    compile options: '-I/usr/local/lib/python3.4/dist-packages/numpy/core/include -I/usr/include/python3.4m -c'
    x86_64-linux-gnu-gcc: scipy/sparse/_csparsetools.c
    In file included from /usr/local/lib/python3.4/dist-packages/numpy/core/include/numpy/ndarraytypes.h:1788:0,
                     from /usr/local/lib/python3.4/dist-packages/numpy/core/include/numpy/ndarrayobject.h:18,
                     from /usr/local/lib/python3.4/dist-packages/numpy/core/include/numpy/arrayobject.h:4,
                     from scipy/sparse/_csparsetools.c:242:
[...]

PR #351, opened 2017-04-18
new interpolator for DDD and other purposes

### LINKED ISSUE #325: can DDD class work with less than three .ddd input files?
https://github.com/pytrip/pytrip/blob/master/pytrip/ddd.py#L135

interpolate.splrep probably requires at least 3 points, otherwise it displays error:
```TypeError: m > k must hold```

grzanka:
Fixes #325 
Partially solving #242

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,7 +98,7 @@ install:
 
 # finally run tox tests
 script:
-  - if [[ $TRAVIS_OS_NAME == "linux" ]]; then travis_wait tox --notest && tox -- -n 8 ; fi
+  - if [[ $TRAVIS_OS_NAME == "linux" ]]; then travis_wait tox --notest && tox -- -n 4 ; fi
 # dropping MacOSX support for Windows, to get faster builds https://github.com/pytrip/pytrip/issues/267
 #  - if [[ $TRAVIS_OS_NAME == "osx" ]]; then travis_wait ./.travis/run_osx.sh ; fi
 # in case all previous commands in script section succeeded, run deploy (if tag is present then upload)
--- a/pytrip/ddd.py
+++ b/pytrip/ddd.py
@@ -50,7 +50,7 @@ class DDD:
         try:
             from scipy import interpolate
         except ImportError as e:
-            logger.error("Please install scipy for you platform to be able to use spline-based interpolation")
+            logger.error("Please install scipy to be able to use spline-based interpolation")
             raise e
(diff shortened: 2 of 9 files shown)


grzanka:
Slowly getting rid of scipy

grzanka:
@nbassler any comments on that ?

it is a helper class to deal with interpolation (linear based on numpy and spline based on scipy) in a pythonic way.

it still needs a bit more polishing and a bit more documentation, but will solve many annoying issues and will help almost to get rid of scipy

nbassler:
Great, also for the insisting windoze users. 
Are you aware of anything higher than linear interpolation is needed?

grzanka:
There are some places where previously cubic spline interpolation was used (i.e. DDD).
On the other hand numpy offers only 1-D linear interpolation. If we would like to interpolate on energies and on depth in DDD then we need some kind of 2-D interpolation. Here only scipy can help (with 2-D bilinear interpolation and 2-D splines).

Here is the list of pytrip features which require scipy:
 - pytrip/ddd.py (2-D spline or linear interpolation)
 - pytrip/utils/bevlet2oer.py (1-D spline interpolation of HLUT data file)

Here only 1-D linear interpolation is used and numpy is enough:
- pytrip/paths.py

nbassler:
I doubt bevlet2oer involves HLUT. It is just a LET->OER translator. Linear interpolation is sufficient here.

grzanka:
My mistake:
 - HLUT interpolation (linear, not spline) is used in DensityCube class in pytrip/paths.py, see https://github.com/pytrip/pytrip/blob/master/pytrip/paths.py#L245 (BTW - is it another cube and another issue with filenames and extensions?)
 - spline interpolation is used in pytrip/utils/bevlet2oer.py to interpolate LET-OER data. I'd be happy to use linear interpolator also there.

nbassler:
Not sure what the HLUT cube is used for, but I think we should follow the file extensions for those which can be loaded or generated by TRiP only, but maybe also have a generic class (.cub?) for anything else a user may produce. (Like Dose*LET / sqrt(phase of moon) + årotons in LHC))

nbassler:
LET->OER interpolation should be replaced anyway with a real model some time.

nbassler:
@grzanka will this also fix #324 ?

grzanka:
Issue #324 should be solved by PR #339

Issue #320, opened 2017-03-13
memory error on 32bit architecture


grzanka:
Appveyor build on 32bit Windows:

```
[00:02:15] ================================== FAILURES ===================================
[00:02:15] _______________________ TestCubeSlicer.test_convert_all _______________________
[00:02:15] 
[00:02:15] self = <tests.test_converters.TestCubeSlicer testMethod=test_convert_all>
[00:02:15] 
[00:02:15]     def test_convert_all(self):
[00:02:15]         working_dir = tempfile.mkdtemp()  # make temp working dir for converter output files
[00:02:15]     
[00:02:15] >       pytrip.utils.cubeslice.main(args=['--data', self.dos, '--ct', self.ctx, '-o', working_dir])
[00:02:15] 
[00:02:15] tests\test_converters.py:243: 
[00:02:15] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[00:02:15] 
[00:02:15] args = Namespace(HUbar=False, csmax=None, ct='tests\\res\\TST003\\tst003000.ctx.gz', data='tests\\res\\TST003\\tst003001.dos.gz', outputdir='C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\tmp8ux0dbhf', sstart=1, sstop=None, verbosity=0)
[00:02:15] 
[00:02:15]     def main(args=sys.argv[1:]):
[00:02:15]         """ The main function for cubeslice.py
[00:02:15]         """
[00:02:15]     
[00:02:15]         # there are some cases when this script is run on systems without DISPLAY variable being set
[00:02:15]         # in such case matplotlib backend has to be explicitly specified
[00:02:15]         # we do it here and not in the top of the file, as interleaving imports with code lines is discouraged
[00:02:15]         import matplotlib
[00:02:15]         matplotlib.use('Agg')
[00:02:15]         import matplotlib.pyplot as plt
[00:02:15]         from matplotlib import colors
[00:02:15]         from matplotlib.ticker import MultipleLocator
[00:02:15]     
[00:02:15]         # parse arguments
[00:02:15]         parser = argparse.ArgumentParser()
[00:02:15]         parser.add_argument("--data", help="data cube(dos, let etc)", type=str, nargs='?')
[00:02:15]         parser.add_argument("--ct", help="CT cube", type=str, nargs='?')
[...]

PR #355, opened 2017-04-24
better job handling on appveyor

grzanka:


nbassler:
#220 is a MR, not an issue?

grzanka:
Should be #320, my mistake

appveyor.yml
```diff
@@ -116,8 +116,11 @@ test_script:
   - cd dist
   - python -c "import os; import pip; pip.main(['install', os.listdir('.')[0]]);"
   - cd ..
+# exclude problematic tests on 32bit platform (see https://github.com/pytrip/pytrip/issues/320)
+  - ps: $exclude_tests = If ($platform -eq "x86") {"test_converters.py"} Else {""}
 # run tests (normal py.test command results in memory error on appveyor, so we run tests file-by-file)
-  - FOR %%I in (tests\test*.py) DO python -m pytest %%I
```
grzanka: This loop was returning True even if one of the tests has failed. I had to replace it with sth more sophisticated - this time Powershell

tests/test_converters.py
```diff
@@ -36,7 +36,7 @@
 import pytrip.utils.bevlet2oer
 import pytrip.utils.rst_plot
 
-from tests.test_base import get_files
+from tests.base import get_files
```
grzanka: test_base was treated as an empty tests and was failing

tests/test_paths.py
```diff
@@ -19,6 +19,31 @@
 """
 TODO: documentation here.
 """
+import unittest
```
grzanka: test_paths was treated as an empty tests and was failing

grzanka:
This PR will not solve #320, but at least any other bugs like this will be reported.

PR #361, opened 2017-04-27
Add a generic volume_histogram function.

### LINKED ISSUE #37: add script: dos+vdx -> DVHs
one command to generate all DVH plots 
optional: export all DVH data to .dat files instead.

### LINKED ISSUE #38: add script: .dosemlet.dos+vdx -> LVHs
same as issue #37 but using LET cubes.

### LINKED ISSUE #360: add generic volume_histogram function
just a suggestion

nbassler:
closes #360 
closes #37 
closes #38

### DIFF
--- a/pytrip/util.py
+++ b/pytrip/util.py
@@ -68,8 +68,8 @@ def volume_histogram(cube, voi=None):
 
     _xrange = (0.0, cube.max()*1.1)
     _hist, x = np.histogram(cube[mask], bins=256, range=_xrange)
-    _fhist = np.fliplr([_hist])[0]  # reverse historgram, so first element is for highest dose
+    _fhist = np.flip(_hist, axis=0)  # reverse historgram, so first element is for highest dose
     _fhist = np.cumsum(_fhist)
-    _hist = np.fliplr([_fhist])[0]  # flip back again to normal
+    _hist = np.flip(_fhist, axis=0)  # flip back again to normal representation
     y = 100.0 * _hist / _hist[0]
     return x[:-1], y  # TODO: think about what a proper representation would be
--- a/pytrip/util.py
+++ b/pytrip/util.py
@@ -69,9 +69,9 @@ def volume_histogram(cube, voi=None, bins=256):
 
     _xrange = (0.0, cube.max()*1.1)
     _hist, x = np.histogram(cube[mask], bins=bins, range=_xrange)
-    _fhist = np.flip(_hist, axis=0)  # reverse historgram, so first element is for highest dose
+    _fhist = _hist[::-1]  # reverse historgram, so first element is for highest dose
     _fhist = np.cumsum(_fhist)
-    _hist = np.flip(_fhist, axis=0)  # flip back again to normal representation
+    _hist = _fhist[::-1]  # flip back again to normal representation
 
     y = 100.0 * _hist / _hist[0]  # volume histograms always plot the right edge of bin, since V(D < x_pos).
     y = np.insert(y, 0, 100.0, axis=0)  # but the leading bin edge is always at V = 100.0%
--- a/pytrip/util.py
+++ b/pytrip/util.py
@@ -20,6 +20,8 @@
 Module with auxilliary functions (mostly internal use).
 """
 
+import numpy as np
+
 
 def get_class_name(item):
     """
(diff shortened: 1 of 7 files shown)


nbassler:
@grzanka a few questions
- wonder if this is a proper way how to calculate e.g. DVHs... It is rather different from the implemented solution in the DosCube and LETCube classes.
- I like when methods have a verb in them, because this distinguishes them from attributes. In that case, this method could be renamed to `make_volume_histogram()` or `get_volume_historgram()` What do you think?

pytrip/util.py
```diff
@@ -43,3 +43,33 @@ def f(x):
...
+    :params cube: a data cube of any shape, e.g. Dos.cube
+    :params voi: optional voi where histogramming will happen.
+    :returns [x],[y]: coordinates ready for plotting. Dose (or LET) along x, Normalized volume along y in %.
+
+    If VOI is not given, it will calculate the histogram for the entire dose cube.
+
+    Providing voi will slow down this function a lot, so if in a loop, it is recommended to do masking
+    i.e. only provide Dos.cube[mask] instead.
+    """
+    import numpy as np
```
grzanka: This should go to top level of the module.

pytrip/util.py
```diff
@@ -43,3 +43,33 @@ def f(x):
...
+
+    if voi is None:
+        mask = np.ones(cube.shape, dtype=bool)
```
grzanka: This is an unnecessary memory overhead. `mask = None` should also work:

```
>>> import numpy as np
>>> x = np.array([[1,2,3],[4,5,6]])
>>> x
array([[1, 2, 3],
       [4, 5, 6]])
>>> mask = None
>>> x[mask]
array([[[1, 2, 3],
        [4, 5, 6]]])
```

pytrip/util.py
```diff
@@ -43,3 +43,33 @@ def f(x):
...
+    else:
+        vcube = voi.get_voi_cube()
+        mask = (vcube == 1000)
+
+    _xrange = (0.0, cube.max()*1.1)
+    _hist, x = np.histogram(cube[mask], bins=256, range=_xrange)
```
grzanka: What about making `bins` a parameter of `volume_histogram` with default value 256 ?

pytrip/util.py
```diff
@@ -43,3 +43,33 @@ def f(x):
...
+    _fhist = np.flip(_hist, axis=0)  # reverse historgram, so first element is for highest dose
+    _fhist = np.cumsum(_fhist)
+    _hist = np.flip(_fhist, axis=0)  # flip back again to normal representation
+    y = 100.0 * _hist / _hist[0]
+    return x[:-1], y  # TODO: think about what a proper representation would be
```
grzanka: Why not the same as `np.histogram`, mainly values + bins edges (of len(values) + 1)?

nbassler: what?

pytrip/util.py
```diff
@@ -43,3 +45,35 @@ def f(x):
...
+        mask = None
+    else:
+        vcube = voi.get_voi_cube()
+        mask = (vcube.cube == 1000)
+
+    _xrange = (0.0, cube.max()*1.1)
+    _hist, x = np.histogram(cube[mask], bins=bins, range=_xrange)
+    _fhist = np.flip(_hist, axis=0)  # reverse historgram, so first element is for highest dose
+    _fhist = np.cumsum(_fhist)
+    _hist = np.flip(_fhist, axis=0)  # flip back again to normal representation
```
grzanka: For python older than 3.4 we need to use numpy < 1.12, see:
https://github.com/pytrip/pytrip/blob/master/requirements.txt

Older version of numpy do not have `flip` method.
Consider using sth like:
```
x = x[::-1]
```

This is the reason travis job fails: https://travis-ci.org/pytrip/pytrip/jobs/226449935

nbassler:
@grzanka should we get rid of the old DVH and LVH routines? May simplify code and less maintenance.

nbassler:
@grzanka you may want take a look at the new dvhplot.py

nbassler:
@grzanka do you have some suggestion how to fix the missing DISPLAY environment case? Setting the matplotlib backend results that the plot is not shown anymore.

grzanka:
@nbassler What you propose in a505b021235f9f5d4d49d1acb52365f9617b81ff looks correct. The idea of separation between pytrip and pytripgui means that the first one is responsible only for logic (or work as a backend) and shouldn't pop-up with any display to the user. It should be solely console application. Generating files is OK, but displaying them not. 

Maybe we should encourage users to use jupyter-notebook and display the images there.

Issue #126, opened 2016-10-12
API change: renaming 


nbassler:
This issue provides a catch all for suggested changes of names

- [ ] vdx.py: VdxCube() -> Vdx() See #97 
- [ ] tripplan.py: no get_*() and set_*() See #118

PR #367, opened 2017-05-03
Adjust API

nbassler:
Related to #126


pytrip/cube.py
```diff
@@ -241,7 +249,7 @@ def create_cube_from_equation(self, equation, center, limits, radial=True):
         y = np.linspace(self.dimx - 0.5, 0.5, self.dimx) * self.pixel_size - center[1]
         xv, yv = np.meshgrid(x, y)
 
-    def load_from_structure(self, voi, preset=0, data_type=np.int16):
+    def set_all_voxels_by_voi(self, voi, preset=0, data_type=np.int16):
```
grzanka: I'd name it `mask_by_voi` as it is clear that we deal with voxels here

tests/test_let.py
```diff
@@ -40,7 +40,7 @@ def setUp(self):
 
     def test_read(self):
         l = LETCube()
-        l.read_trip_data_file(self.let001)
+        l._read_trip_data_file(self.let001)
```
grzanka: This is a private method and shouldn't be used here.

nbassler:
@grzanka better?

grzanka:
@nbassler much better

PR #371, opened 2017-06-12
Fix RBE issue #370

### LINKED ISSUE #370: RBE models: add handler for zero dose, and fix another bug in models
Carabe models return RBE values below 1.0 even in SOBP, right now.

nbassler:
Fixes #370

### DIFF
--- a/pytrip/models/proton.py
+++ b/pytrip/models/proton.py
@@ -113,19 +113,20 @@ def rbe_mcnamara(dose, let, abx):
     return rbe
 
 
-def _rbe_apx(dose, apx, sbpx, abx):
+def _rbe_apx(dose, apx, sbpx, abx, dzero=0.0):
     """
     :params dose: proton dose      [Gy]
     :params apx: alpha_p / alpha_x [dimensionless] RBE_max = ap/ax when (dose -> 0 Gy)
     :params sbpx: beta_p / beta_x  [dimensionless] RBE_min = sqrt(bp/bx) when (dose -> inf Gy)
     :params abx: alpha_x / beta_x  [Gy^-1]
+    :params dzero: what to return in case of dose is zero (which would cause division by zero)
     """
 
     _rbe = 1.0 / (2.0 * dose)
     if hasattr(_rbe, '__iter__'):
-        _rbe[_rbe == np.inf] = 0
+        _rbe[_rbe == np.inf] = dzero
     else:
         if _rbe == np.inf:
-            return 0.0
+            return dzero
     _rbe *= (np.sqrt(abx*abx + 4*abx*apx*dose + 4*sbpx*sbpx * dose*dose) - abx)
     return _rbe
--- a/pytrip/models/proton.py
+++ b/pytrip/models/proton.py
@@ -51,7 +51,7 @@ def rbe_carabe(dose, let, abx):
     TODO: handle Cube() class directly
 
     :params dose: physical proton dose in [Gy]
-    :params let: LET in [keV/um]
+    :params let: LETd in [keV/um] (protons only)
     :params abx: alpha_x / beta_x [Gy^-1]
 
     :returns: RBE for the given parameters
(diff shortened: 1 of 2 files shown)


pytrip/models/proton.py
```diff
@@ -123,5 +123,7 @@ def _rbe_apx(dose, apx, bpx, abx):
     :params abx: alpha_x / beta_x  [Gy^-1]
     """
 
-    _rbe = (np.sqrt(abx*abx + 4*abx*apx*dose + 4*bpx * dose*dose) - abx) / (2 * dose)
+    _rbe = 1.0 / (2.0 * dose)
+    _rbe[_rbe == np.inf] = 0
+    _rbe *= (np.sqrt(abx*abx + 4*abx*apx*dose + 4*bpx*bpx * dose*dose) - abx)
```
nbassler: bpx*bpx in third term

pytrip/models/proton.py
```diff
@@ -108,20 +107,21 @@ def rbe_mcnamara(dose, let, abx):
...
     :params dose: proton dose      [Gy]
     :params apx: alpha_p / alpha_x [dimensionless] RBE_max = ap/ax when (dose -> 0 Gy)
-    :params bpx: beta_p / beta_x   [dimensionless] RBE_min = sqrt(bp/bx) when (dose -> inf Gy)
+    :params sbpx: beta_p / beta_x  [dimensionless] RBE_min = sqrt(bp/bx) when (dose -> inf Gy)
     :params abx: alpha_x / beta_x  [Gy^-1]
     """
 
-    _rbe = (np.sqrt(abx*abx + 4*abx*apx*dose + 4*bpx * dose*dose) - abx) / (2 * dose)
+    _rbe = 1.0 / (2.0 * dose)
+    _rbe[_rbe == np.inf] = 0
```
grzanka: Isn't it better to set `_rbe[_rbe == np.inf] = np.nan` ?

`np.nan` is nicely handed in numpy and helps to avoid problems with division by zero when one calculates `D_ion = D_gamma / RBE`

nbassler: Working with arrays containing np.nan throws tons of errors when plotting, histogramming etc, and causes me to hate pytrip, since things just get more complicated and time consuming to work with. RBE = 0 is totally valid imho. Dose is zero anyway here.

pytrip/models/proton.py
```diff
@@ -108,20 +107,25 @@ def rbe_mcnamara(dose, let, abx):
...
     :params apx: alpha_p / alpha_x [dimensionless] RBE_max = ap/ax when (dose -> 0 Gy)
-    :params bpx: beta_p / beta_x   [dimensionless] RBE_min = sqrt(bp/bx) when (dose -> inf Gy)
+    :params sbpx: beta_p / beta_x  [dimensionless] RBE_min = sqrt(bp/bx) when (dose -> inf Gy)
     :params abx: alpha_x / beta_x  [Gy^-1]
     """
 
-    _rbe = (np.sqrt(abx*abx + 4*abx*apx*dose + 4*bpx * dose*dose) - abx) / (2 * dose)
+    _rbe = 1.0 / (2.0 * dose)
+    if hasattr(_rbe, '__iter__'):
+        _rbe[_rbe == np.inf] = 0
```
grzanka: I propose to add a paremeter to `_rbe_apx` (and other methods) named `outside_domain` with default value zero.
This parameter should be returned if RBE model cannot calculate a resonable value.

In some cases I'd prefer to use `np.inf` than zero and such option would give me some flexibility.

nbassler: I wonder what those cases would be look like, where you need RBE = np.inf instead of 0.

grzanka: Following case, assuming we have a region where RBE=0 (due to model limitation), we have some reference radiation dose D_phys = X, and calculate biological dose as 
D_phys / RBE. We obtain `np.inf`.

In this case if RBE would be np.inf, then D_biol would be 0.

Giving user an option (with default setting of RBE=0) would allow them more flexibility in this case.

PR #383, opened 2017-07-26
typo fixed

### LINKED ISSUE #382: Typo in executer code
It is crashing when str method is called, see
https://github.com/pytrip/pytrip/blob/master/pytrip/tripexecuter/execute.py#L111

```
        out += "|   STDOUT prefix               : '{:s}'\n".format(self.logfile_prefiz_stdout)
```

It should be prefix, not prefiz

grzanka:
Fixes #382


grzanka:
@nbassler ready for review

PR #421, opened 2018-02-21
support for newer (>1.0) version of pydicom package

grzanka:
which require new way of importing

### DIFF
--- a/pytrip/ctx.py
+++ b/pytrip/ctx.py
@@ -118,7 +118,9 @@ class CtxCube(Cube):
             if ds.SOPInstanceUID.startswith('2.25.'):
                 # UUID based UIDs
                 # example: 2.25.137355362850316362338405159557803441718
-                uuid_part_str = ds.SOPInstanceUID[len('2.25.'):]  # extract last part of UID, as string
+                uuid_part_str = ds.SOPInstanceUID[len('2.25.'):len('2.25.') + 32]  # extract 32bit fragment of
+                # last part of UID, as string
+
                 uuid_object = uuid.UUID(int=int(uuid_part_str))   # convert to UID object, to be able to manipulate it
                 uuid_list = list(uuid_object.fields)              # get list of fields, to be able to edit it
                 uuid_list[-1] = i + 1  # replace clock_seq part of uuid with sequential number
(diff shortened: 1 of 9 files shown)


grzanka:
Most of the problem fixed, but one funny remain.

```
..
======================================================================
ERROR: test_create_voi_2 (__main__.TestVdx)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_vdx.py", line 220, in test_create_voi_2
    v = create_voi_from_cube(c, name="cube4")
  File "/home/grzanka/PycharmProjects/pytrip/pytrip/vdx.py", line 501, in create_voi_from_cube
    import matplotlib._cntr as cntr
ModuleNotFoundError: No module named 'matplotlib._cntr'
```

I've checked on my local PC that matplotlib 2.2.0rc1 released on 14.02.2017 is missing a hidded feature of contour lines extraction as described here: https://stackoverflow.com/questions/18304722/python-find-contour-lines-from-matplotlib-pyplot-contour

```
→ python3                                                                                                                            ~/PycharmProjects/pytrip
Python 3.6.4 (default, Jan  5 2018, 02:35:40) 
[GCC 7.2.1 20171224] on linux
>>> from matplotlib import pyplot as plt
>>> from matplotlib import _cntr as cntr
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name '_cntr'
>>> from matplotlib import __version__
>>> __version__
'2.2.0rc1'
```

grzanka:
It looks this feature was removed ion Sep 22, 2017:  https://github.com/matplotlib/matplotlib/commit/e95572c68529c9c2af743805d7e514b944759646

and will not be available starting from matplotlib 2.2.0

pytrip/ctx.py
```diff
@@ -118,7 +118,9 @@ def create_dicom(self):
             if ds.SOPInstanceUID.startswith('2.25.'):
```
nbassler: maybe this part deserves more explanation why it is needed

grzanka: I agree, but anyway DICOM related stuff needs to be refactored

PR #438, opened 2018-04-25
VolHist related updates

nbassler:
- add write method to VolHist
- add deprecation warnings to old DVH method
- add logger to let.py


nbassler:
this is not urgent.

PR #448, opened 2018-05-02
new features

### LINKED ISSUE #447: better plotting in spc2pdf
- [ ] plotting with dots instead with lines
- [ ] PDF metadata

grzanka:
Fixes #447

### DIFF
--- a/pytrip/utils/spc2pdf.py
+++ b/pytrip/utils/spc2pdf.py
@@ -20,9 +20,12 @@
 """
 Script for inspecting SPC files and producing PDF with summary plots
 """
-import sys
-import logging
 import argparse
+import datetime
+import getpass
+import logging
+import math
+import sys
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


pytrip/utils/spc2pdf.py
```diff
@@ -159,7 +178,10 @@ def main(args=sys.argv[1:]):
                         energy_steps = data.energy[m3]
                         fluence = data.fluence[m3]
                         if np.any(m3):
-                            ax.plot(energy_steps, fluence, label="Z = {:d}, depth = {:3.3f} [cm]".format(z, depth_step))
+                            ax.plot(energy_steps, fluence,
+                                    linestyle=linestyle,
+                                    marker=marker,
+                                    label="Z = {:d}, depth = {:3.3f} [cm]".format(z, depth_step))
```
nbassler: do we need to repeat the depth here, if it is already set in the title? We can save some legend space.

grzanka: There are two depths: one requested by user, which goes into title, and another one for which spectrum is shown (from the one available in SPC file) which goes into legend. They may differ.  See for example:

image

nbassler: So there are two different kinds of depth. Need better naming scheme then. If you show that plot above to anyone, they will be confused..

grzanka: What do you think about following:
image

?

grzanka: Fixed in c31aa24

PR #453, opened 2018-05-07
Fix volhist issues

### LINKED ISSUE #451: VolHist: does not warn properly if invalid Voi.
```
>>> dvh = VolHist(d,voi)
/home/bassler/.local/lib/python3.6/site-packages/pytrip/volhist.py:118: RuntimeWarning: invalid value encountered in true_divide
  y = 100.0 * _hist / _hist[0]  # volume histograms always plot the right edge of bin, since V(D < x_pos).
>>> pt.__version__
'2.3.0'
```
Using TST003 from pytrip/tests/res:

```
import pytrip as pt

c = pt.CtxCube()
c.read("tst003000.ctx.gz")

v = pt.VdxCube(c)
v.read("tst003000.vdx")

d = pt.DosCube()
d.read("tst003001.dos.gz")

voi = v.vois[1]

from pytrip.volhist import VolHist
dvh = VolHist(d,voi)
```

### LINKED ISSUE #452: VolHist: does not work with some LetCubes
```
  File "/home/bassler/.local/lib/python3.6/site-packages/pytrip/volhist.py", line 43, in __init__
    self.target_dose = cube.target_dose  # optional target dose scaling factor
AttributeError: 'LETCube' object has no attribute 'target_dose'
Aborted
```
which makes sense when you look at the code. LETCubes may not have the target_dose attribute.

nbassler:
Fix #451 
Fix #452

### DIFF
--- a/pytrip/let.py
+++ b/pytrip/let.py
@@ -49,6 +49,7 @@ class LETCube(Cube):
 
     def __init__(self, cube=None):
         super(LETCube, self).__init__(cube)
+        self.type = "LET"
         self.let_type = None
 
     def get_max(self):
--- a/pytrip/volhist.py
+++ b/pytrip/volhist.py
@@ -27,6 +27,7 @@ class VolHist:
     """
     Volume histogram class
     """
+
     def __init__(self, cube, voi=None, target_dose=None):
         """
         :params Cube cube: either LETCube, DosCube or similar object.
@@ -101,6 +102,8 @@ class VolHist:
 
         If VOI is not given, it will calculate the histogram for the entire dose cube.
 
+        If VOI is a point or has no contents, all y values are set to 0.
+
         Providing voi will slow down this function a lot, so if in a loop, it is recommended to do masking
         i.e. only provide Dos.cube[mask] instead.
         """
@@ -110,6 +113,12 @@ class VolHist:
         else:
             vcube = voi.get_voi_cube()
             mask = (vcube.cube == 1000)
+
+        if True not in mask:
+            logging.warning("Given VOI has no extend and contains no voxels.")
+            x = np.arange(100)
+            y = np.zeros(100)
+            return x, y
 
         _xrange = (0.0, data_cube.max() * 1.1)
         _hist, x = np.histogram(data_cube[mask], bins=bins, range=_xrange)
--- a/pytrip/volhist.py
+++ b/pytrip/volhist.py
@@ -20,6 +20,7 @@
 This module provides the volume histogram class.
 """
 import logging
+import warnings
 import numpy as np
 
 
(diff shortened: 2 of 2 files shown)


pytrip/volhist.py
```diff
@@ -109,6 +115,13 @@ def volume_histogram(data_cube, voi=None, bins=256):
             vcube = voi.get_voi_cube()
             mask = (vcube.cube == 1000)
 
+        if True not in mask:
```
grzanka: Better `if not mask.any():`

pytrip/volhist.py
```diff
@@ -109,6 +115,13 @@ def volume_histogram(data_cube, voi=None, bins=256):
...
+            warnings.warn("Given VOI has no extend and contains no voxels.",
+                          UserWarning)
+            x = np.arange(100, dtype=np.float64)
+            y = np.zeros(100, dtype=np.float64)
+            return x, y
```
grzanka: I think we should return here `None,None` and handle it properly on the other (GUI) side

pytrip/let.py
```diff
@@ -49,6 +49,7 @@ class LETCube(Cube):
 
     def __init__(self, cube=None):
         super(LETCube, self).__init__(cube)
+        self.type = "LET"
```
grzanka: What is that ?

pytrip/volhist.py
grzanka: Also tests fails with:

```
>       if True not in mask:
E       TypeError: argument of type 'NoneType' is not iterable
```

pytrip/let.py
nbassler: DosCube has one, so why should LETCube not have one too?

grzanka: I forgot to remove it from initial import, it wasn't used if I am correct.

nbassler: uh, I intended to use it for the GUI

grzanka: why not a class name / type ?

nbassler:
```
import pytrip as pt
if isinstance(obj, pytrip.DosCube):
     foobar()
```
vs.
```
if obj.type =='DOS';
    foobar()
```

you can save one line of code, and I just found the latter solution a bit leaner and more convenient, even if less pythonic.

pytrip/volhist.py
```diff
@@ -109,6 +115,11 @@ def volume_histogram(data_cube, voi=None, bins=256):
             vcube = voi.get_voi_cube()
             mask = (vcube.cube == 1000)
 
+        if not mask.any():
```
grzanka:
```
        if voi is None:
            mask = None
        else:
            vcube = voi.get_voi_cube()
            mask = (vcube.cube == 1000)
    
>       if not mask.any():
E       AttributeError: 'NoneType' object has no attribute 'any'
```

mask can be also None

nbassler:
New issue.

pytrip/let.py
nbassler: lets put that in a new issue

PR #463, opened 2018-05-24
Possible fix for #460, missing z-tables.

### LINKED ISSUE #460: missing ztable when reading out ctx file
So, I generate a .ctx and .hed from a DICOM file set with dicom2trip. The resulting .hed file has a ztable.
I then create the simplest of scripts which does nothing but load in the ctx (and .hed at the same time) and then write it out again (and at the same time checks if ztable is present or not):

```
my_ctx = pt.CtxCube()
my_ctx.read(ctx_path)
print(my_ctx.z_table)
my_ctx.write("testwriteout.ctx")
``` 

The testwriteout.hed written out with the code above does not have a ztable!
`print(my_ctx.z_table)` returns `False` but should be `True`, if I am not mistaken. Therefore I think the error happens when the .hed is loaded in with the .read comand and not doing the writing out with the .write command.

nbassler:
Closes #460.

### DIFF
--- a/pytrip/cube.py
+++ b/pytrip/cube.py
@@ -135,7 +135,7 @@ class Cube(object):
             # unique for each CT slice
             self._ct_sop_instance_uid = uid.generate_uid(prefix=None)
 
-            self.z_table = False  # list of slice#,pos(mm),thickness(mm),tilt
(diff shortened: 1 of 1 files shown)


nbassler:
Moved down for cosmetics. There should only be a z_table if the previous command succeeds.

There is still some ambiguity, since we both may have a z-table in .hed and slice positions from .vdx. There are no rules what will happen if they are not in sync.

Issue #459, opened 2018-05-22
fix missing iterations command


TokePrintz:
In the function  `_make_exec_opt` in _plan.py_ the iteration argument should also be written out. The iteration number is set to 500 as default in L110 but not further used.

It's an easy fix, just add ` opt += "iter({:d})".format(self.iterations)` somewhere before L534.

nbassler:
feel free to commit a patch and setup a MR :)

grzanka:
PR, not MR...

nbassler:
old gitlab habit..

TokePrintz:
Speakin as one who does not yet have any old habits on either gitlab nor github, was exactly do we mean when we say "MR" and "PR"? My guess would be that MR is merge request and PR then maybe patch request but was is the difference. I planned just to create a new branch, commit and do a merge request as I did all the other times?

nbassler:
Stackoverflow for the win

https://stackoverflow.com/questions/22199432/pull-request-vs-merge-request

TokePrintz:
Always! Without Stackoverflow, I would still be in the stage of hitting my fellow human beings on the head with a bone.

...I was just not 100% sure what PR meant in this context...

nbassler:
closed via #464

PR #464, opened 2018-05-24
Fix missing iterations command, #459

TokePrintz:
suggest for #459

### DIFF
--- a/pytrip/tripexecuter/plan.py
+++ b/pytrip/tripexecuter/plan.py
@@ -531,6 +531,8 @@ class Plan(object):
         # TODO: sanity check numbers
         opt += " geps({:.2e})".format(self.geps)
         opt += " eps({:.2e})".format(self.eps)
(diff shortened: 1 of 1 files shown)


pytrip/tripexecuter/plan.py
```diff
@@ -531,6 +531,8 @@ def _make_exec_opt(self):
         # TODO: sanity check numbers
         opt += " geps({:.2e})".format(self.geps)
         opt += " eps({:.2e})".format(self.eps)
+        opt += "iter({:d})".format(self.iterations)
```
nbassler: You will need a leading space before iter(). I doubt TRiP parser would accept this...

TokePrintz: Ofc! My bad!

TokePrintz: New code has been committed

PR #470, opened 2018-06-28
Concept of KernelModel

### LINKED ISSUE #458: tripexecutor: add KernelModel object
makes GUI easier and more flexible. Something like:
```
import logging

logger = logging.getLogger(__name__)


class KernelModel(object):
    """
    Kernel data, holding all necessary data for a certain projectile
    """

    projectile_z = 0
    projectilz_a = 0
    projectile_iupac = ""  # chemical symbol (parsed by TRiP), "H", "He", "C", "O", "Ne" ....
    projectile_name = ""  # cleartext name "Protons", "Carbon-12 ions", "Antiprotons", ...
    rifi_name = ""  # name of ripple filter, e.g. "3 mm Rifi", for documentation only
    rifi_thickness = ""
    ddd_path = ""  # "/local/trip/data/ddd/*"
    spc_path = ""  # "/local/trip/data/spc/*"
    sis_table = ""  # if none, it will be autogenerated
    comment = ""  # arbitrary text

    def __init__(self):
        pass
```
This object can then also be added to Plan class. Field will then select from a list of kernels in plans.

nbassler:
this will not break compability.

Closes #458


pytrip/tripexecuter/kernel.py
```diff
@@ -0,0 +1,69 @@
...
+    rifi_name = ""  # name of ripple filter, e.g. "3 mm Rifi", for documentation only
+    rifi_thickness = ""
+    ddd_path = ""  # "/local/trip/data/ddd/*"
+    spc_path = ""  # "/local/trip/data/spc/*"
+    sis_path = ""  # if none, it will be autogenerated
+    comment = ""  # arbitrary text
+
+    def __init__(self):
+        """
+        :param symb: iupac symbol of projectile.
```
grzanka: Comment not related to the method.

pytrip/tripexecuter/projectile.py
```diff
@@ -0,0 +1,81 @@
...
+        if symb in self._projectile_defaults:
+            self.z = self._projectile_defaults[symb][0]
+            self.a = self._projectile_defaults[symb][1]
+
+        self.name = name
+        if z > 0:
+            self.z = z
+
+        if a > 0:
+            self.a = a
```
grzanka: Give a warning if setting `z` and `a` wasn't succesfult. Then 0,0 will be set as default and strange things may happen.

nbassler:
This should also prepare for future multi-ion optimization.

examples/example02_trip98.py
```diff
@@ -50,16 +50,23 @@
...
+mykernel.sis_path = "/home/bassler/TRiP98/base/DATA/SIS/12C.sis"
+mykernel.rifi_thickness = 3.0  # 3 mm ripple filter. (Only for documentaiton, will not affect dose optimization.)
+mykernel.rifi_name = "GSI_1D_3mm"  # Additional free text for documentation.
+mykernel.comment = "Carbon-12 ions with 3 mm 1D Ripple Filter"
+
+# Ok, we have the Contours, the CT cube and dose kernels ready. Next we must prepare a plan.
 # We may choose any basename for the patient. All output files will be named using
 # this basename.
-plan = pte.Plan(basename=patient_name)
+plan = pte.Plan(basename=patient_name, kernels=[mykernel])
```
grzanka: Using a tuple  `kernels=(mykernel,)` would be more elegant (and less error prone) - we don't wish to modify it in function call.

tests/test_exec_250.py
```diff
@@ -0,0 +1,116 @@
...
+
+        logger.debug("Load VdxCube {:s}".format(self.vdx_path))
+        v = pt.VdxCube(c)
+        v.read(self.vdx_path)
+
+        print(v.voi_names())
+
+        kernel = pte.KernelModel("C12 Ions RiFi 3MM")
+        kernel.projectile = pte.Projectile("C")
+        kernel.ddd_path = "~/TRiP98/base/DATA/DDD/12C/RF3MM/*"
```
grzanka: We also ship pytrip to windows users. Making tests with `~` may be a bit confusing for them

examples/example02_trip98.py
nbassler: IMHO it looks less readable, due to that trailing comma.

nbassler:
would also like to hear @TokePrintz opinion

TokePrintz:
Sry, has been afk the whole day. Promise to look at it this evening.

nbassler:
@TokePrintz Yes. the idea is to have all the kernel parameters in a object.

Then it is much easier to mess around with separate beam models. I.e. you can setup a `mykernel` for each of your ripple filters. See the updated version of example 02.

PR #474, opened 2018-07-04
VDX writer: Avoid repeating single points in a slice 

### LINKED ISSUE #472: Generating spherical PTV: At first and last slice, same line written out twice
I generate .ctx and .vdx files with a spherical target as contour in the latter using the following code snip:

```
    # create the CT cube with constant HU value corresponding to lung
    my_ctx = pt.CtxCube()
    my_ctx.basename = patient_name  # give a name to the ctx file to be generated doing planning   
    my_ctx.create_empty_cube(value=-400, dimx=80, dimy=80, dimz=80, pixel_size=1.0, slice_distance=1.0, slice_offset=0.0)  #TODO:ctsteps(1,1,1) is probably too fine!

    # create a sphere as PTV and add it to VdxCube
    my_ptv = pt.vdx.create_sphere(my_ctx, 'PTV', [40,40,depth], 20.0)
    my_vdx = pt.VdxCube()
    my_vdx.basename = patient_name  # give a name to the vdx file to be generated doing planning
    my_vdx.add_voi(my_ptv)
```

but I have discovered that the first and last slice in the .vdx file for the contour has the line with the coordinates repeated:

```
slice 0
slice_in_frame 0.000
thickness 2.000 reference start_pos -1.000 stop_pos 1.000
number_of_contours 1
contour 0
internal false
number_of_points 1
 64.0000 64.0000 0.0000 0.0000 0.0000 0.0000
 64.0000 64.0000 0.0000 0.0000 0.0000 0.0000
```

which results in an error from TRiP98 in the planning phase. By manually deleting the repeated line in the first and last slice, I can run the plan using the generated .exec file.

The error must occur in vdx.py L606-L607, which handles the r2 == 0.0 case, e.g. the slices of the outmost parts of the sphere, where the contour is just a single point (it is this single point that is written out twice), but the code is so short and simple that I really cannot see where something could go wrong...

nbassler:
Closes #472

### DIFF
--- a/pytrip/vdx.py
+++ b/pytrip/vdx.py
@@ -1550,7 +1550,7 @@ class Contour:
                                                          0, 0, 0)
 
         # repeat the first point, to close the contour, if needed
-        if self.contour_closed:
+        if self.contour_closed and len(self.contour) > 1:
             out += " %.4f %.4f %.4f %.4f %.4f %.4f\n" % (self.contour[0][0] - self.cube.xoffset,
                                                          self.contour[0][1] - self.cube.yoffset,
                                                          self.contour[0][2],


nbassler:
@TokePrintz  can you test this?

If it resolves, mark it as "Ready" and feel free to merge.

TokePrintz:
@grzanka shoud I just merge or do you also want to review?

PR #475, opened 2018-07-04
Small fix for KernelModel

### LINKED ISSUE #473: minor KernelModel issues
```
    rifi_thickness = ""
```
should be float in mm

`__str__()` method needs fixing

```
    print(obj)
  File "/home/bassler/.local/lib/python3.6/site-packages/pytrip/tripexecuter/kernel.py", line 47, in __str__
    return self._print()
  File "/home/bassler/.local/lib/python3.6/site-packages/pytrip/tripexecuter/kernel.py", line 56, in _print
    out += "|  RiFi Name                         : {:s}\n".format(str(self.rifi.name))
AttributeError: 'KernelModel' object has no attribute 'rifi'

```

nbassler:
Closes #473


TokePrintz:
Aaaah, I see now. The bug is that there has to be a float value given.
 Okay, I guess we have to choose something and then maybe 0.0 is the best :-)

PR #501, opened 2019-10-25
Remove dependency on old matplotlib

grzanka:
Simply a copy of C file with contouring algorithm from matplotlib 2.1.2 was made.
This algorithm at the moment is only used in `create_voi_from_cube` method in `vdx` class
(which "Creates a new VOI which holds the contours following an isodose lines.")

### DIFF
--- a/pytrip/lib/cntr.c
+++ b/pytrip/lib/cntr.c
@@ -1,0 +1,2038 @@
+/* -*- mode: c; c-basic-offset: 4 -*- */
+
+/*
+  cntr.c
+  General purpose contour tracer for quadrilateral meshes.
+  Handles single level contours, or region between a pair of levels.
+
+    The routines that do all the work, as well as the explanatory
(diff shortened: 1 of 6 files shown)


pytrip/lib/cntr.c
```diff
@@ -0,0 +1,2038 @@
...
+  Handles single level contours, or region between a pair of levels.
+
+    The routines that do all the work, as well as the explanatory
+    comments, came from gcntr.c, part of the GIST package. The
+    original mpl interface was also based on GIST.  The present
+    interface uses parts of the original, but places them in
+    the entirely different framework of a Python type.  It
+    was written by following the Python "Extending and Embedding"
+    tutorial.
+
```
nbassler: Consider adding a link where it is from?

grzanka: added

PR #506, opened 2020-01-30
voi_cube doesn't has max() method

ljelen:


ljelen:
Cherypick commit from #510

PR #513, opened 2020-02-01
adaptation of C extension to the new API

### LINKED ISSUE #512: Rewrite C extension using newest numpy C API
In numpy 1.7 (released February 2013, https://pypi.org/project/numpy/1.7.0/) direct access to PyArrayObject was deprecated, see https://numpy.org/devdocs/release/1.7.0-notes.html

> Direct access to the fields of PyArrayObject* has been deprecated. Direct access has been recommended against for many releases. Expect similar deprecations for PyArray_Descr* and other core objects in the future as preparation for NumPy 2.0.

Newer numpy version offer different API, see https://docs.scipy.org/doc/numpy/reference/c-api.array.html

For example, code using pre 1.7 API looks like this:
```
double ** pyvector_to_array(PyArrayObject *arrayin)  {
    int i, rows;
    double * array;
    double **out;

    array = (double *) arrayin->data;  /* pointer to arrayin data as double */
    rows = arrayin->dimensions[0];
    out = (double **)malloc(sizeof(double *)*rows);
    for(i = 0; i < rows; i++)
    {
        out[i] = (double *)malloc(sizeof(double)*2);
        out[i][0] = array[2*i];
        out[i][1] = array[2*i+1];
    }
    return out;
}
```

With new API one could use i.e. `npy_intp PyArray_DIM(PyArrayObject* arr, int n)` to get number of rows and `void *PyArray_DATA(PyArrayObject *arr)` to get data.  

One could question if it is necessary at all to convert from numpy arrays to C arrays

grzanka:
Fixes #512


grzanka:
Summary of changes:
 - core code moved from `filter_point.c` (meaningless name now) to `core.c`. This code is being used in python part and covered by tests. Extension name `pytriplib` is maintained
 - unused and untested code moved to `unused.c` which is compiled into separate extension called `unused`
 - data access to numpy arrays done now using `PyArray_GETPTR?` methods. It replaces direct access to underlying data pointer and ensure correct operation for C and Fortran data alignment.
 - Python C list API used in the places where C extension method needs to return a list
 - many comments added

PR #515, opened 2020-02-02
parse_exec should be part of public API

### LINKED ISSUE #514: Using proteced method outside class
https://github.com/pytrip/pytrip/blob/8edbe5b677e1f5085d1bc9af22a7fcd386d060de/pytrip/tripexecuter/plan.py#L307

parse_exec should not be protected

ljelen:
closes #514


grzanka:
@ljelen fix PEP8 / flake problems

Issue #533, opened 2020-05-19
Newest (1.19) numpy drops support for python3.5


grzanka:
See i.e. https://travis-ci.com/github/pytrip/pytripgui/jobs/337079290

grzanka:
Apart from requirements.txt we should also consider updating install_requires in setup.py. Take a look at https://hynek.me/articles/conditional-python-dependencies/

grzanka:
Fixed in https://github.com/pytrip/pytrip/pull/535/files

PR #534, opened 2020-05-19
Special numpy version for python 3.5

grzanka:
See #533


grzanka:
Seems like travis fails due to quai outage: https://status.quay.io/incidents/kw2627bsdwd9

PR #551, opened 2020-09-16
idea for simplified RST manipulation #550

nbassler:


nbassler:
@grzanka what do you think?

PR #553, opened 2020-12-17
Sending stdout while trip98 is working

ljelen:
This change is needed by gui to properly display progress

https://github.com/pytrip/pytripgui/issues/292


grzanka (approved):
Looks OK

PR #568, opened 2021-04-03
Execute trip remotely

ljelen:
Remove code duplication
Connection with private key (un)protected works
Connection with username/password works

related to:
https://github.com/pytrip/pytripgui/pull/357


grzanka (approved):
Looks ok

PR #579, opened 2021-06-12
Add validation for new VOI - can't exceed patient cube boundaries

Hewwho:


grzanka (changes requested):
During last meeting we decided to investigate `get_min_max` method implemented in `Voi` class and all its members:
https://github.com/pytrip/pytrip/blob/master/pytrip/vdx.py#L1194

Due to rounding of the values it may give different results that the check based on arithmetical condition of radius, center and cube dimension. Moreover `get_min_max` may be used for whatever shape of VOI.

I'd like to have both method implemented in `pytrip` and used in `pytripgui`, but before proceeding further some detail checks need to be performed to see if a method based on `get_min_max` gives the same result as the arithmetic method (at least for spheres).

grzanka:
See also https://github.com/pytrip/pytrip/issues/580

grzanka (changes requested):
Resolve comments

pytrip/vdx.py
```diff
@@ -1210,6 +1210,14 @@ def get_min_max(self):
         self.temp_max = temp_max
         return temp_min, temp_max
 
+    def is_fully_contained(self):
+        min_pos, max_pos = self.get_min_max()
```
grzanka, line 1214: Replace indexing with plain variables. Use 3 variables for each condition, along X, Y and Z

grzanka (changes requested):

pytrip/vdx.py
```diff
@@ -1210,6 +1210,30 @@ def get_min_max(self):
...
 
+    def is_fully_contained(self):
+        try:
+            [min_pos_x, min_pos_y, min_pos_z], [max_pos_x, max_pos_y, max_pos_z] = self.get_min_max()
+        except TypeError:
+            # get_min_max can return NoneTypes if a VOI is located outside of the patient
+            return False
+
+        if [min_pos_x, min_pos_y, min_pos_z] == [max_pos_x, max_pos_y, max_pos_z]:
+            # get_min_max can return equal min and max values if a VOI is partially contained in the patient
```
grzanka, line 1221: keep in mind that VOI can also be a single point, what in that case ?

pytrip/vdx.py
```diff
@@ -1210,6 +1210,30 @@ def get_min_max(self):
...
+            return False
+
+        return self.is_x_contained(min_pos_x, max_pos_x) and \
+               self.is_y_contained(min_pos_y, max_pos_y) and \
+               self.is_z_contained(min_pos_z, max_pos_z)
+
+    def is_x_contained(self, min_pos, max_pos):
```
grzanka, line 1228: These methods looks like something that shoun't be exposed in class API, consider making them "private".
In order to do so, rename them to `_is_x_contained`

grzanka (changes requested):

pytrip/vdx.py
```diff
@@ -531,7 +531,7 @@ def create_voi_from_cube(cube, name, value=100):
         points = np.zeros((len(contour[0]), 3))
         points[:, 0:2] = contour[0] * cube.pixel_size
 
-        points[:, 2] = i * cube.slice_distance
+        points[:, 2] = i * cube.slice_distance + cube.zoffset
```
grzanka, line 534: As a test:
1. Make programmatically in python an empty patient with sperical VOI with negative Z offset, save the files (*.ctx, *.hed, *.vdx)
2. In another program read the generated files (*.ctx, *.hed, *.vdx), check if Z offset is negative and if the data structures are the same as in step 1.

pytrip/vdx.py
```diff
@@ -556,8 +556,8 @@ def create_cylinder(cube, name, center, radius, depth):
     p = list(zip(center[0] + radius * np.cos(t), center[1] + radius * np.sin(t)))
 
     for i in range(0, cube.dimz):
-        z = i * cube.slice_distance
-        if center[2] - depth / 2 <= z <= center[2] + depth / 2:
+        z = i * cube.slice_distance + cube.zoffset
```
grzanka, line 559: Make proper documentation how the VDX data is being handled, where offsets are added/subtracted and what is saved/read to/from the VDX files.
Investigate also so called z_table.

pytrip/vdx.py
grzanka: Withdraw changes related to adding VOI with z offset, will be added later in new issue

grzanka:
@Hewwho please rebase as well

grzanka:
@Hewwho which pytrip version should I use to test this feature ?
Now I get:
```
Traceback (most recent call last):
  File "/home/grzanka/PycharmProjects/pytripgui/pytripgui/add_vois_vc/add_single_voi_vc/add_single_voi_cont.py", line 27, in _save_and_exit
    if self._validate_voi():
  File "/home/grzanka/PycharmProjects/pytripgui/pytripgui/add_vois_vc/add_single_voi_vc/add_single_voi_cont.py", line 62, in _validate_voi
    if not voi.is_fully_contained():
AttributeError: 'Voi' object has no attribute 'is_fully_contained'
```

pytrip/vdx.py
```diff
@@ -1287,6 +1278,26 @@ def get_min_max(self):
         self.temp_max = temp_max
         return temp_min, temp_max
 
+    def is_fully_contained(self):
```
grzanka, line 1281: Add docstring explaining what this method is doing.
Take some inspiration from 
https://github.com/pytrip/pytrip/pull/589/files#diff-3db93b3f0fd4220f8fa3abdfe45db89baae38ab57a33c3109f2eaab2f3c5eccfR809

grzanka (changes requested):

grzanka (approved):
Looks OK

grzanka:
Please squash merge once tests are ok

PR #586, opened 2021-08-18
Fix DICOM data element keywords for export

Yoanka:
Some DICOM data element keywords were incompatible with the current standard and thus not supported by PyDICOM:
http://dicom.nema.org/medical/dicom/current/output/html/part06.html

This caused the following warnings and a corrupted export from PytripGUI:
image


grzanka (approved):
Looks OK

Issue #588, opened 2021-08-23
3D data indexing


grzanka:
It seems that right now pytrip stores 3D cubes, indexing them with Z,Y,X order.
For example, typing `ctx_cube.cube[0,:,:]` would give us Y,X plane intersection.
I suspect most of the users would expect in such case rather Y,Z plane intersection

grzanka:
I think we should discuss moving into X,Y,Z indexing, as it seems more natural for users and more compatible with DICOM format.

grzanka:
Moreover pytrip itself looks inconsisten, as parts written in C seems to assume X,Y,Z convention, see:
```
        if(plane == 2) // sagittal, projection onto YZ
        {
            first_point_x = *((double*)PyArray_GETPTR2(vec_slice, i, 0));
            second_point_x = *((double*)PyArray_GETPTR2(vec_slice, i+1, 0));
            if((first_point_x >= depth && second_point_x < depth) || (second_point_x >= depth && first_point_x < depth))
            {
                first_point_y = *((double*)PyArray_GETPTR2(vec_slice, i, 1));
                second_point_y = *((double*)PyArray_GETPTR2(vec_slice, i+1, 1));
                first_point_z = *((double*)PyArray_GETPTR2(vec_slice, i, 2));
                second_point_z = *((double*)PyArray_GETPTR2(vec_slice, i+1, 2));
```
https://github.com/pytrip/pytrip/blob/master/pytrip/lib/core.c#L676

PR #589, opened 2021-08-23
documenting code

grzanka:
Related to #588

### DIFF
--- a/pytrip/vdx.py
+++ b/pytrip/vdx.py
@@ -643,7 +643,7 @@ class Voi:
         out += "   Voi\n"
         out += "----------------------------------------------------------------------------\n"
         out += "|   Name                                : '{:s}'\n".format(self.name)
-        out += "|   Is concated                         : {:s}\n".format(str(self.is_concated))
+        out += "|   Is concatenated                     : {:s}\n".format(str(self.is_concated))
         out += "|   Type                                : {:d}\n".format(self.type)
         out += "|   Number of slices in VOI             : {:d}\n".format(len(self.slices))
         out += "|   Color 0xRGB                         : #{:s}{:s}{:s}\n".format(hex(self.color[0].strip('0x')),
@@ -719,7 +719,7 @@ class Voi:
 
     def get_3d_polygon(self):
         """ Returns a list of points rendering a 3D polygon of this VOI, which is stored in
-        sefl.polygon3d. If this attribute does not exist, create it.
+        self.polygon3d. If this attribute does not exist, create it.
         """
         if not hasattr(self, "polygon3d"):
             self.concat_to_3d_polygon()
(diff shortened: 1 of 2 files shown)


grzanka:
With huge help of @rozchlastywacz

pytrip/vdx.py
```diff
@@ -791,33 +791,74 @@ def get_2d_projection_on_basis(self, basis, offset=None):
 
     def get_2d_slice(self, plane, depth):
         """
-        Gets a 2d Slice object from the contour in either sagittal or coronal plane.
+        Gets a 2D Slice object from the contour in either sagittal or coronal plane.
```
nbassler, line 794: It would be great if the docstring would list how the contour points are ordered for each plane (Z,X) (X,Z) (Z,Y) or (Y,Z).
(Also great if we can mark the funtions if they can handle non-equidistant slices, and non-continous slices, which may appear in DICOM.)

pytrip/vdx.py
```diff
@@ -791,33 +791,74 @@ def get_2d_projection_on_basis(self, basis, offset=None):
...
+
+            # close the contour
+            contour.append(contour[0])
+
+            # create a slice object and add a contour data
             s = Slice(cube=self.cube)
-            s.add_contour(Contour(points1, cube=self.cube))
+            s.add_contour(Contour(contour, cube=self.cube))
 
         return s
```
nbassler, line 863: otherwise nice with extensive comments :-)

pytrip/vdx.py
```diff
@@ -791,33 +791,74 @@ def get_2d_projection_on_basis(self, basis, offset=None):
 
     def get_2d_slice(self, plane, depth):
         """
-        Gets a 2d Slice object from the contour in either sagittal or coronal plane.
+        Gets a 2D Slice object from the contour in either sagittal or coronal plane.
         Contours will be concatenated.
 
         :param int plane: either self.sagittal or self.coronal
```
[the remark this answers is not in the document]
grzanka: I agree
we are planning to implement new version of the algorithm with support for concave contours.
I will be supported then
https://github.com/pytrip/pytrip/issues/590

pytrip/vdx.py
grzanka: Good idea, we will add it !
This is important, and I suspect in some places still broken, see https://github.com/pytrip/pytrip/issues/593

pytrip/vdx.py
```diff
@@ -791,33 +791,74 @@ def get_2d_projection_on_basis(self, basis, offset=None):
...
         """
-        self.concat_contour()
-        points1 = []
-        points2 = []
-        for _slice in self.slices:  # TODO: slices must be sorted first, but wouldnt they always be ?
+        # concat_contour() merges all contours to one contour, as in TRiP98 standard
+        self.concat_contour()  # TODO: this is modifying current Voi, which is not nice, refactor it
+
+        # variables to store points that have same depth, but different X or Y coordinate depending on plane type
+        # lower chain contains points with higher coordinate value (X or Y)
```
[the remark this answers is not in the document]
grzanka: If a contour is crossed by a straight line, then there may be two intersection points.
These two points are then added to one of two lists: 
 - list of points with higher X (for coronal plane) or Y (for sagittal plane)
 - list of points with lower X (for coronal plane) or Y (for sagittal plane)
 
 One list will form "upper part" of the resulting contour, another - "lower part". 
Finally we merge "upper part" with "lower part", but to maintain correct sequence of point we go though lower part in reverse way.

PR #642, opened 2021-11-18
Python 3.10 support

grzanka:
This is a massive PR aiming at adding support for python 3.10. However, on the way to this goal, several other improvements had to be added:
 - removal of unittest framework from test suite in favour of pytest. Previously both were mixed. Switching to pytest gives cleaner architecture.
 - removal of tox which was not fully exploited and provided an additional level of complexity. pytest and github actions matrix tests provide the necessary coverage for all python versions
 - migrating all requirements into `setup.py`. Now `requirements.txt` contains single instruction to install pytrip package in editable mode. This dramatically simplifies dependency management.


tests/test_exec.py
```diff
@@ -19,93 +19,70 @@
 """
 TODO: documentation here.
```
rozchlastywacz, line 20: :)

rozchlastywacz (approved):
Pytest with test methods looks a lot cleaner and more readable than tox with test classes.
Alt Text

tests/test_exec.py
grzanka: There is more of that, will come in different PR

.github/workflows/main.yml
```diff
@@ -110,13 +118,16 @@ jobs:
...
-      uses: RalfG/python-wheels-manylinux-build@v0.3.3-manylinux1_x86_64
+      uses: RalfG/python-wheels-manylinux-build@v0.4.0-manylinux1_x86_64
       with:
        # Python versions to target, space-separated
        python-versions: "cp35-cp35m cp36-cp36m cp37-cp37m cp38-cp38 cp39-cp39"
 
+    - name: Python wheels manylinux build
+      uses: RalfG/python-wheels-manylinux-build@v0.4.0-manylinux2010_x86_64
+      with:
+       python-versions: "cp310-cp310"  # Python versions to target, space-separated
```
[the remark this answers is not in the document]
grzanka: First one is for manylinux1 standard which has support only until python 3.9, second one is for manylinux2010 with python 3.10.

PR #666, opened 2021-12-21
reduce remote results size

michal367:
remote execution was sending back all files including input files (e.g. ctx, which is big)


grzanka (approved):
LGTM

PR #667, opened 2021-12-22
fix loading bio dose on execute

michal367:


pytrip/tripexecuter/execute.py
```diff
@@ -411,16 +411,16 @@ def _finish(self, plan):
             path = os.path.join(plan.temp_dir, file_name)
             self._log("Reading {:s} file".format(file_name))
             if ".phys.dos" in file_name:
-                ctx_cube = DosCube()
+                dos_cube = DosCube()
```
grzanka, line 414: I reccoment to switch to utils class:
https://github.com/pytrip/pytrip/blob/a7583edaa5261ed01a48397cfab7947a5aa5d800/pytrip/util.py#L159
instead of checking `if ".phys.dos" in file_name:`

Two options:
1. Use `is_valid_cubeclass`
2. Use `suffix` attribute and compare against list of allowed_suffixes (i.e.https://github.com/pytrip/pytrip/blob/a7583edaa5261ed01a48397cfab7947a5aa5d800/pytrip/dos.py#L54)

grzanka (approved):
LGTM

PR #671, opened 2022-01-05
Hotfix for deprecated dicom tags in dos.py

Yoanka:
image

All tags were also verified as up to date with DICOM 2021e (current version).


grzanka (approved):
LGTM

PR #728, opened 2023-09-18
test stuff

nbassler:


grzanka:
Can be replaced by https://github.com/pytrip/pytrip/pull/732

